PR #39, opened 2017-07-20
Bugfix for probability prediction with caret

grayskripko:
caret:::predict.train expects colnames of a predicted probability dataset match to target levels. 
At the moment, catboost.caret$prob for binary classification returns a matrix with colnames: c("", "prediction")
This is the reason of the error in caret:::predict.train:
```
Error in `[.data.frame`(out, , obsLevels, drop = FALSE) : 
  undefined columns selected
```


a4tunado:
Good job thanks for contribution!

PR #59, opened 2017-07-24
catboost.caret looping functionality

hadjipantelis:
Added functionality so `catboost.caret` can loop through the number of boosting iterations when training.
The following search grid would:
```R
grid <- expand.grid(depth = c(2, 3), learning_rate = 0.1, , border_count = c(255),
                    iterations = c(1:20), l2_leaf_reg = c(1e-6,1e-3), rsm = 0.95)
```
would result into 80 (`2 * 20 * 2`) models being fitted. The new `loop` functionality allows fitting just 4 (`2 * 2`) models as we fit models with 20 trees each and then the `catboost.caret$loop` makes the necessary looping to generate predictions for the relevant/lower `tree_count_limit` arguments (`1:19`).

### DIFF
--- a/catboost/R-package/R/catboost.caret.R
+++ b/catboost/R-package/R/catboost.caret.R
@@ -80,6 +80,7 @@ catboost.caret$fit <- function(x, y, wts, param, lev, last, weights, classProbs,
 #' @param preProc: preprcess data option
 #' @param submodels: only used with the loop element
 catboost.caret$predict <- function(modelFit, newdata, preProc = NULL, submodels = NULL) {
+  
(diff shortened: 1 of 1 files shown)


hadjipantelis:
Hi @a4tunado @exprmntr,

Is there a problem with this PR? It does provide functionality that `catboost.caret` does not have; the `loop` functionality is very helpful as it allows testing multiple runs in one go. Please let me know if you have any particular concerns you would like me to fix.

annaveronika:
Hi, sorry for the delay, we are looking on it now, we will answer soon.

PR #98, opened 2017-09-01
disable scrollZoom and enable image save

newbfg:


newbfg:
I hereby agree to the terms of the CLA available at: https://yandex.ru/legal/cla/?lang=ru

PR #106, opened 2017-09-19
Typo in method: model.feature_importance_

math-and-data:
Should call print(model.feature_importance_) not print(model.feature_importances_)


annaveronika:
The field is  called feature_importances_, so the initial code is correct, this change will break it.

PR #110, opened 2017-09-21
Increased number of threads

gamers5a:
Parameter CB_THREAD_LIMIT increased from 32 to 128 for more complete utilization of multiprocessor systems.

I hereby agree to the terms of the CLA available at: [https://yandex.ru/legal/cla/?lang=en].


annaveronika:
For your case with 100 cores you could recompile the library with updated limit and use it. We will update the limit to 128 at some point, but I am not sure it is a very common case for now.

Issue #156, opened 2017-10-31
Python CatBoostRegressor does not predict with small inputs


avidale:
When applying the fitted model to matrix 1*1, it gives an error:
`CatboostError: Input data has invalid shape or empty: (1, 1). Must be 2 dimensional`

The problem is that `catboost.core.Pool._check_data_empty` checks for input size are too strict. **This method should be rewritten**.
As a result, the following code throws an error in the last line:
```
import catboost
import numpy as np
x_train3 = np.array([[1,2,3,], [2,3,4], [3,4,5]])
x_train1 = np.array([[1], [2], [3]])
y_train = np.array([1,2,3])
x_test3_2 = np.array([[4,5,6], [5,6,7]])
x_test3_1 = np.array([[4,5,6,]])
x_test1_2 = np.array([[4], [5]])
x_test1_1 = np.array([[4]])
model3 = catboost.CatBoostRegressor().fit(x_train3, y_train)
model1 = catboost.CatBoostRegressor().fit(x_train1, y_train)
print(model3.predict(x_test3_2)) # OK
print(model3.predict(x_test3_1)) # OK
print(model1.predict(x_test1_2)) # OK
print(model1.predict(x_test1_1)) # Throws an error!
```

It is already mentioned [on stackowerflow](https://stackoverflow.com/questions/47039915/catboost-understanding-conversion-of-categorical-values).

I propose to replace the check in the line 194 of `core.py`
`if not len(data_shape) == 2 or not sum(data_shape) > 2:`
by 
`if not len(data_shape) == 2 or not min(data_shape) > 0:`

annaveronika:
would you make a pull request with your change?

annaveronika:
Thank you for the fix!

PR #157, opened 2017-10-31
Allow 1*1 input matrix

avidale:
This single-line update allows to use 1*1 input matrices when predicting with Python package, as described in the issue #156 .

As a new contributor, I hereby agree to the terms of the CLA available at: https://yandex.ru/legal/cla/?lang=en.

### DIFF
--- a/catboost/python-package/catboost/core.py
+++ b/catboost/python-package/catboost/core.py
@@ -191,7 +191,7 @@ class Pool(_PoolBase):
                     data_shape = tuple(data_shape + tuple([len(data[0])]))
                 else:
                     data_shape = tuple(data_shape + tuple([1]))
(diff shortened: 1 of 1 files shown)


catboost/python-package/catboost/core.py
```diff
@@ -191,7 +191,7 @@ def _check_data_empty(self, data):
                     data_shape = tuple(data_shape + tuple([len(data[0])]))
                 else:
                     data_shape = tuple(data_shape + tuple([1]))
-            if not len(data_shape) == 2 or not sum(data_shape) > 2:
+            if not len(data_shape) == 2 or not min(data_shape) > 2:
```
annaveronika: you mean 0?

avidale: Yes, I meant 0, as in the issue. Corrected it.

PR #213, opened 2018-01-03
Misspelling of useful

yyq:
It looks like a missspelling


yyq:
Only changed a few words in the tutorial document. But that two CI failed. Someone maintainer will take a look ?

PR #248, opened 2018-02-11
use ipywidgets 7.x+ syntax by default

newbfg:
ipywidgets 7.x+ are required


newbfg:
@annaveronika we need update our docs:
`ipywidgets 7.x+ are required`

JustM57:
Home PC still need to recreate notebooks 0.6.2(cat)

PR #399, opened 2018-06-27
Added an option to allow negative weights

kazeevn:
Adds an option to allow negative document weights.

Negative weights are not insanity, they are commonly used in High-Energy Physics in the following way. Take a classification dataset. Add some noise documents, that are from neither class. Then, by some way independent of the features present in the dataset, for each document compute the probability of it being noise. Then, there is a way to compute weights from these probabilities in such a way, that, when the dataset is weighted with those weights, the feature's distribution will be that of the original dataset before the noise addition. More in the paper: https://arxiv.org/abs/physics/0402083

Example pool: https://cernbox.cern.ch/index.php/s/XK3ZuNlGCg1Dtc3

### DIFF
--- a/catboost/app/bind_options.cpp
+++ b/catboost/app/bind_options.cpp
@@ -570,6 +570,12 @@ void ParseCommandLine(int argc, const char* argv[],
             (*plainJsonPtr)["allow_const_label"] = true;
         });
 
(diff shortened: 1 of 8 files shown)


annaveronika:
We rely on the optimized function to be convex, if we want the training to converge. With negative weights we cannot guarantee that training will converge. So we decided to not support negative weights. You can patch the code and use the patched code if you really want them to be present. But we cannot guarantee that there will be a good result.

PR #455, opened 2018-08-26
save settings state during a CV learning process

newbfg:


annaveronika:
@arcadia-devtools Ship it!

PR #486, opened 2018-10-02
Handle nulls and OutOfMemoryErrors slightly better in the JNI interface

mikekap:
I hereby agree to the terms of the CLA available at: https://yandex.ru/legal/cla/?lang=en .

### DIFF
--- a/catboost/jvm-packages/catboost4j-prediction/src/native/ai_catboost_CatBoostJNIImpl.cpp
+++ b/catboost/jvm-packages/catboost4j-prediction/src/native/ai_catboost_CatBoostJNIImpl.cpp
@@ -93,7 +93,7 @@ static void HashCatFeatures(
         // NOTE: instead of C-style cast `dynamic_cast` should be used, but compiler complains that
         // `_jobject` is not a polymorphic type
         const auto jcatFeature = (jstring)jenv->GetObjectArrayElement(jcatFeatures, i);
-        CB_ENSURE(jcatFeature, "got null array element");
+        CB_ENSURE(jenv->IsSameObject(jcatFeature, NULL) == JNI_FALSE, "got null array element");
         Y_SCOPE_EXIT(jenv, jcatFeature) {
             jenv->DeleteLocalRef(jcatFeature);
         };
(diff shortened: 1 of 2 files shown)


catboost/jvm-packages/catboost4j-prediction/src/native/ai_catboost_CatBoostJNIImpl.cpp
```diff
@@ -92,8 +93,13 @@ static void HashCatFeatures(
         // NOTE: instead of C-style cast `dynamic_cast` should be used, but compiler complains that
         // `_jobject` is not a polymorphic type
         const auto jcatFeature = (jstring)jenv->GetObjectArrayElement(jcatFeatures, i);
+        CB_ENSURE(jcatFeature, "got null array element");
```
yazevnul: This is a wrong way to check for `null`, you should do `jenv->IsSameObject(jcatFeature, NULL) == JNI_FALSE`.

catboost/jvm-packages/catboost4j-prediction/src/native/ai_catboost_CatBoostJNIImpl.cpp
```diff
@@ -92,8 +93,13 @@ static void HashCatFeatures(
...
+        Y_SCOPE_EXIT(jenv, jcatFeature) {
+            jenv->DeleteLocalRef(jcatFeature);
```
yazevnul: Are you sure this is really necessary? Isn't JNI suppose to discard all local references on exit from native method?

catboost/jvm-packages/catboost4j-prediction/src/native/ai_catboost_CatBoostJNIImpl.cpp
mikekap: I was reading https://books.google.com/books?id=8M3F_sSSvWkC&lpg=PA66&ots=Qn07SiJ2fk&pg=PA66#v=onepage&q&f=false and it mentioned that you can compare by equality unless you have weak references. I think that only applies if you store weak references in your JNI class - in either case it wouldn't apply here since all the references are strong. I'm happy to change it if that's the style you prefer though.

catboost/jvm-packages/catboost4j-prediction/src/native/ai_catboost_CatBoostJNIImpl.cpp
mikekap: It's not strictly necessary, but I think there's a maximum on the number of local references allowed per JNI call. I figured any kind of loop with user-supplied data can overflow that limit (I think it's 512 by default?) That said my only evidence of this is stories like https://github.com/assimp/assimp/issues/712

yazevnul:
@mikekap, any updates?

mikekap:
Sorry for the delay - I addressed your comments. Let me know if the changes make sense.

yazevnul:
@arcadia-devtools Ship it!

arcadia-devtools:
@yazevnul, internal review request created: 575340

PR #557, opened 2018-11-29
[WIP] Snapshot to model

ilivans:
Convert snapshot saved during training to applicable model file.

Works only for numeric features so far. I suppose that categorical features support is possible only in presence of training dataset. The option is to be added soon.

Other model file formats are going to be supported as well.

- [x]  Load snapshot from scratch
- [x]  Convert snapshot into model
- [x]  Export model
- [x]  Test
- [ ]  Categorical features support
- [ ]  Multiple model file formats support

### DIFF
--- a/catboost/app/mode_snapshot_to_model.cpp
+++ b/catboost/app/mode_snapshot_to_model.cpp
@@ -1,0 +1,84 @@
+#include "modes.h"
+#include "bind_options.h"
+
+#include <catboost/libs/algo/learn_context.h>
+#include <catboost/libs/algo/split.h>
+#include <catboost/libs/helpers/progress_helper.h>
+#include <catboost/libs/model/model_build_helper.h>
+#include <library/getopt/small/last_getopt.h>
+
+
+struct TLoadSnapshotParams {
+    TString SnapshotPath;
+    TString ModelPath;
+
+    void BindParserOpts(NLastGetopt::TOpts& parser) {
+        parser.AddLongOption('s', "snapshot-file", "path to load snapshot from")
+                .StoreResult(&SnapshotPath)
+                .RequiredArgument("PATH");
+        parser.AddLongOption('m', "model-file", "path to save model at")
+                .StoreResult(&ModelPath)
+                .RequiredArgument("PATH");
+    }
+};
+
+
+TLearnProgress LoadSnapshot(const TString& snapshotPath) {
+    CB_ENSURE(NFs::Exists(snapshotPath), "Snapshot file doesn't exist: " << snapshotPath);
+    try {
+        TLearnProgress learnProgress;
+        TProgressHelper(ToString(ETaskType::CPU)).CheckedLoad(snapshotPath, [&](TIFStream* in) {
+            TRestorableFastRng64 rand(0);
+            ::Load(in, rand);
+            learnProgress.Load(in, true);
+        });
+        CATBOOST_INFO_LOG << "Snapshot is loaded from: " << snapshotPath << Endl;
+        return learnProgress;
+    } catch (...) {
+        CATBOOST_ERROR_LOG << "Can't load progress from snapshot file: " << snapshotPath << Endl;
+        throw;
+    }
+}
+
+
+TObliviousTrees BuildObliviousTrees(const TLearnProgress& learnProgress) {
+    try {
+        TObliviousTreeBuilder builder(learnProgress.FloatFeatures, learnProgress.CatFeatures,
+                                      learnProgress.ApproxDimension);
+        for (size_t treeId = 0; treeId < learnProgress.TreeStruct.size(); ++treeId) {
+            TVector <TModelSplit> modelSplits;
+            for (const auto &split : learnProgress.TreeStruct[treeId].Splits) {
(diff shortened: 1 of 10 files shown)


catboost/libs/algo/split.h
```diff
@@ -109,7 +110,7 @@ struct TSplit : public TSplitCandidate {
     inline void Load(IInputStream* s) {
         ::LoadMany(s, static_cast<TSplitCandidate&>(*this), BinBorder);
     }
-    TModelSplit GetModelSplit(const TLearnContext& ctx, const TDataset& learnData) const;
+    TModelSplit GetModelSplit(const TLearnProgress& learnProgress, const TCtrHelper& ctrHelper, const TDataset& learnData) const;
```
persiyanov: Why not just get `ctx.LearnProgress` & `ctx.CtrsHelper` inside the function? Why changing the signature?

catboost/pytest/test.py
```diff
@@ -5642,3 +5642,55 @@ def test_broken_dsv_format(dataset_name, loss_function, has_pairs, has_group_wei
 
     with pytest.raises(yatest.common.ExecutionError):
         yatest.common.execute(cmd)
+
+
+@pytest.mark.parametrize('pool', ['higgs'])
```
persiyanov: If you pass only one value for `pool`, `pytest.mark.parametrize` is not needed then, I suppose :)

catboost/libs/algo/split.h
ilivans: Context object (ctx) is used during training (it's filled with some stuff based on training dataset at least), but we don't have one in our mode, we still want to use this method though. So that's why I reduced the dependency space of the method.

catboost/pytest/test.py
ilivans: True. It was supposed to work for another pool too, but that's still in progress. Thanks for pointing out.

catboost/pytest/test.py
```diff
@@ -5642,3 +5642,55 @@ def test_broken_dsv_format(dataset_name, loss_function, has_pairs, has_group_wei
...
+        '--eval-file', output_eval_path,
+        '-m', output_model_path,
+        '--model-format', 'CatboostBinary',
+        '--snapshot-file', snapshot_path
+    )
+    yatest.common.execute(cmd)
+    convert_cmd = (
+        CATBOOST_PATH,
+        'snapshot-to-model',
+        '-s', snapshot_path,
```
annaveronika: I suggest to change -s to --cpu_snapshot, since there is also gpu snapshot which has a different format.

catboost/app/mode_snapshot_to_model.cpp
```diff
@@ -0,0 +1,84 @@
...
+#include "bind_options.h"
+
+#include <catboost/libs/algo/learn_context.h>
+#include <catboost/libs/algo/split.h>
+#include <catboost/libs/helpers/progress_helper.h>
+#include <catboost/libs/model/model_build_helper.h>
+#include <library/getopt/small/last_getopt.h>
+
+
+struct TLoadSnapshotParams {
```
annaveronika: please use unnamed namespace here

catboost/app/mode_snapshot_to_model.cpp
```diff
@@ -0,0 +1,84 @@
...
+                .StoreResult(&SnapshotPath)
+                .RequiredArgument("PATH");
+        parser.AddLongOption('m', "model-file", "path to save model at")
+                .StoreResult(&ModelPath)
+                .RequiredArgument("PATH");
+    }
+};
+
+
+TLearnProgress LoadSnapshot(const TString& snapshotPath) {
```
annaveronika: please use static functions here

catboost/app/mode_snapshot_to_model.cpp
```diff
@@ -0,0 +1,84 @@
...
+}
+
+
+TObliviousTrees BuildObliviousTrees(const TLearnProgress& learnProgress) {
+    try {
+        TObliviousTreeBuilder builder(learnProgress.FloatFeatures, learnProgress.CatFeatures,
+                                      learnProgress.ApproxDimension);
+        for (size_t treeId = 0; treeId < learnProgress.TreeStruct.size(); ++treeId) {
+            TVector <TModelSplit> modelSplits;
+            for (const auto &split : learnProgress.TreeStruct[treeId].Splits) {
```
annaveronika: Please check out the style guide for things like space after TVector and position of &. Here it should be:
TVector<TModelSplit>
const auto& split

catboost/app/modes.h
```diff
@@ -7,3 +7,4 @@ int mode_metadata(int argc, const char* argv[]);
 int mode_run_worker(int argc, const char* argv[]);
 int mode_roc(int argc, const char* argv[]);
 int mode_model_sum(int argc, const char* argv[]);
+int mode_snapshot_to_model(int argc, const char* argv[]);
```
annaveronika: The most important thing is to have it in python. So I would suggest adding this to catboost.load_model(format='CpuSnapshot')

annaveronika:
We are currently planning our next release. For now the planned date is Friday. It would be super nice to have this pr in.

annaveronika:
Closing because of inactivity. Feel free to create one more pr, if you're ready.

PR #589, opened 2018-12-18
UPD: shap calculation benchmark

canorbal:
Update benchmark for shap values calculation.
ADD: asymptotic analysis, time measurements for data preprocessing
CHANGED: range of test size
DELETED: useless info about GPU


canorbal:
@annaveronika yes, it was my mistake
The final PR can be found [here](https://github.com/catboost/catboost/pull/590)

PR #600, opened 2018-12-26
Remove avg lines/points with disabled std

newbfg:
Remove avg lines/points with disabled std in Jupyter Notebook


annaveronika:
@arcadia-devtools Ship it!

arcadia-devtools:
@annaveronika, internal review request created: 656475

PR #623, opened 2019-01-17
Fix string missing format

kevinyen-oath:
Now the error it prints is 
```CatboostError: type of {}_feature_names[{}]: expected str, found {}```
since the string is not properly formatted. Without looking into code, people will be confused.

_I hereby agree to the terms of the CLA available at: https://yandex.ru/legal/cla/?lang=en_


annaveronika:
@arcadia-devtools Ship it!

arcadia-devtools:
@annaveronika, internal review request created: 671549

PR #652, opened 2019-02-02
Create compare method

Drakon5999:
Task 4

Compare function:
Draw train and eval errors in Jupyter notebook for two models. 
If metrics already calculated you can call it like model1.compare(model2)
Else you should add all parameters that need for eval_metrics function 

--
I hereby agree to the terms of the CLA available at https://yandex.ru/legal/cla/?lang=en.


arcadia-devtools:
@annaveronika, internal review request created: 701547

PR #645, opened 2019-02-02
[python-package] Implement __eq__ for CatBoost* classes.

daskol:
WIP

PR #654, opened 2019-02-02
[python-package] Implement __eq__ for CatBoost* classes

daskol:
Another one attempt to contribute something (see #645 and #650).

### DIFF
--- a/catboost/python-package/ut/medium/test.py
+++ b/catboost/python-package/ut/medium/test.py
@@ -19,7 +19,6 @@ from catboost import (
     cv,
     sum_models,
     train,)
-from catboost.datasets import titanic
 from catboost.eval.catboost_evaluation import CatboostEvaluation
 from catboost.utils import eval_metric, create_cd, get_roc_curve, select_threshold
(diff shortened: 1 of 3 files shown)


annaveronika:
@arcadia-devtools Ship it!

arcadia-devtools:
@annaveronika, internal review request created: 701573

catboost/python-package/ut/medium/test.py
```diff
@@ -3664,3 +3664,34 @@ def test_eval_set_with_no_target_with_eval_metric(task_type):
     )
     with pytest.raises(CatboostError):
         model.fit(train_pool, eval_set=eval_set_pool)
+
+
+def test_model_comparison():
+    def fit_model(value):
```
andrey-khropov: After changing to adult pool this function no longer depends on value so both model1 and model2 will be the same. I propose make 'iterations' a parameter and check 3 vs 5 iterations, for example.

daskol: Done.

PR #655, opened 2019-02-02
WIP: Add SampleId as a main name for sample id column in a column description file

daskol:


annaveronika:
please agree to cla, see examples in other pr-s

annaveronika:
@arcadia-devtools Ship it!

annaveronika:
Please add a test that shows that both DocId and SampleId work

catboost/libs/column_description/cd_parser.cpp
```diff
@@ -70,6 +70,9 @@ namespace {
                 if (type == "Target") {
                     type = "Label";
                 }
+                if (type == "DocId") {
```
andrey-khropov: The same logic should be added where output columns are parsed in eval_result.cpp

arcadia-devtools:
@annaveronika, internal review request created: 701585

catboost/libs/column_description/cd_parser.cpp
daskol: Hm, it is tricky. I was expected that `column_descriptor/` is the only place where string names of column types are used. I think it would be better to pass `TVector<TColumn>` instead of `TVector<TString>` to `OutputEvalResultToFile()`.

andrey-khropov: > I think it would be better to pass TVector<TColumn> instead of TVector<TString> to OutputEvalResultToFile().

I agree but this is more involved refactoring outside of this particular change.

annaveronika:
@arcadia-devtools Ship it!

arcadia-devtools:
@annaveronika, internal review request created: 709727

annaveronika:
@daskol You need to agree to our cla (see examples in other pr-s), otherwise we will not be able to merge your pr.

annaveronika:
@daskol Do you plan to proceed with the pr?

andrey-khropov:
@arcadia-devtools Ship it!

arcadia-devtools:
@andrey-khropov, internal review request created: 737216

PR #805, opened 2019-04-26
Added guards to avoid segmentation faults when _fit is called with None parameters

### LINKED ISSUE #804: Python: Segmentation fault when `y` or `eval_set` parameters are None
Problem: Segmentation fault when 
catboost version: 0.14.2
Operating System: Mac OSX Sierra 10.12.6 and Linux 3.10.0-693.17.1.el7.
CPU: x86_64

Example code that makes the failure:
```python

import catboost as cb
import pandas as pd
import numpy as np


data = pd.DataFrame({
              'color': np.random.randint(1, 10, 100),
              'article_id': np.random.randint(100, 125, 100),
              'weight': np.random.rand(100),
              'y': np.arange(100) ** 2
})
y = data['y']
data = data.drop('y', axis=1)

cr = cb.CatBoostRegressor(loss_function='RMSE', eval_metric="RMSE")
fit_params={'use_best_model': True, 'cat_features': [0, 1], 'verbose': False}

# Segmentation fault here
cr.fit(data, y, eval_set=(None, None), **fit_params)

# Segmentation fault here as well
cr.fit(None, y, **fit_params)
```
Looking at the code, it seems that some guards are missing. When `y` is None there's an error but the message is cryptic.  Having not `None` paramters might be obvious when using for experiments but when is used inside a software this might happen and having a clear error is better than a seg fault (or a weird message). 

I will submit a small PR for solving this.

mfcabrera:
Fixes #804 

I don't see tests for the python package itself. It would be a good idea to have some for the python wrapper as well, to test the API behaviour. I tested my changes manually making sure a `CatBoostError` exception is raised when certain parameters (IMHO some exceptions here should be ValueError, to be more specific). I also ran the tests using  `ya make -t -A`. 

I hereby agree to the terms of the CLA available at:  https://yandex.ru/legal/cla/?lang=en

### DIFF
--- a/catboost/python-package/catboost/core.py
+++ b/catboost/python-package/catboost/core.py
@@ -1,4 +1,3 @@
-
 import sys
 from copy import deepcopy
 from six import iteritems, string_types, integer_types
@@ -1108,6 +1107,12 @@ class CatBoost(_CatBoostBase):
              column_description, verbose_eval, metric_period, silent, early_stopping_rounds,
              save_snapshot, snapshot_file, snapshot_interval):
 
+        if X is None:
+            raise CatBoostError("X must not be None")
+
+        if y is None:
+            raise CatBoostError("y must not be None")
+
         params = deepcopy(self._init_params)
         if params is None:
             params = {}
@@ -1184,7 +1189,10 @@ class CatBoost(_CatBoostBase):
             elif isinstance(eval_set, tuple):
                 if len(eval_set) != 2:
                     raise CatBoostError("Invalid shape of 'eval_set': {}, must be (X, y).".format(str(tuple(type(_) for _ in eval_set))))
+                if eval_set[0] is None or eval_set[1] is None:
+                    raise CatBoostError("'eval_set' tuple contains at least one None value")
                 eval_sets.append(Pool(eval_set[0], eval_set[1], cat_features=train_pool.get_cat_feature_indices()))
+
                 eval_total_row_count += eval_sets[-1].num_row()
                 if eval_sets[-1].num_row() == 0:
                     raise CatBoostError("Empty 'eval_set' in tuple")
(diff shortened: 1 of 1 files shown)


david-waterworth:
@mfcabrera does this cover the case where X is a Pool(), in which case y should be None?

mfcabrera:
@david-waterworth I've changed the condition. I wasn't aware of it. Thanks :)

kizill:
@arcadia-devtools Ship it!

arcadia-devtools:
CLA already signed

arcadia-devtools:
@kizill, internal review request created: [808715](https://a.yandex-team.ru/review/808715)

arcadia-devtools:
Check [808715](https://a.yandex-team.ru/review/808715) failed: 4 test(s) broken

catboost/python-package/catboost/core.py
```diff
@@ -1108,6 +1107,12 @@ def _fit(self, X, y, cat_features, pairs, sample_weight, group_id, group_weight,
              column_description, verbose_eval, metric_period, silent, early_stopping_rounds,
              save_snapshot, snapshot_file, snapshot_interval):
 
+        if X is None:
+            raise CatBoostError("X must not be None")
+
+        if y is None and not isinstance(X, Pool):
```
kizill: X could also be a string with dataset file path

mfcabrera: @kizill thanks, I saw that but on the documentation of `y` it is not described:
``` y : list or numpy.array or pandas.DataFrame or pandas.Series, optional (default=None)
            Labels, 1 dimensional array like.
            Use only if X is not catboost.Pool.```

If I read this, I believe that  whenever y is None then X should be an instance of Pool. I think the doc shoudl say that. Should I update the docstring as well as the `isinstance` condition?

mfcabrera: @kizill I have added the additional condition.

kizill:
@arcadia-devtools Ship it!

arcadia-devtools:
@kizill, internal review request updated: [808715](https://a.yandex-team.ru/review/808715)

arcadia-devtools:
Check [808715](https://a.yandex-team.ru/review/808715) passed

PR #1047, opened 2019-10-25
Use environment marker to specify enum34 dependency

tomzx:
Use environment markers ([PEP-0508](https://www.python.org/dev/peps/pep-0508/)) for typing_extensions to specify
that enum34 is only required if python < 3.4. Some dependency management
tools (including poetry) only work reliable with environment markers
instead of conditional checks in setup.py.

### DIFF
--- a/catboost/python-package/setup.py
+++ b/catboost/python-package/setup.py
@@ -27,10 +27,8 @@ install_requires=[
     'scipy',
     'plotly',
     'six',
+    'enum34; python_version < "3.4"'
 ],
-
-if sys.version_info < (3, 4):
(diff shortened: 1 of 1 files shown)


kizill:
@arcadia-devtools Ship it!

arcadia-devtools:
CLA already signed

arcadia-devtools:
@kizill, internal review request created: [1004715](https://a.yandex-team.ru/review/1004715)

arcadia-devtools:
Check [1004715](https://a.yandex-team.ru/review/1004715) passed

PR #1180, opened 2020-02-19
add UserPerObjMetric as a regression objective

camrongodbout:
## PR Objective:
Allow for users to use a custom defined metric as regression objective

## What Changed:
* Added a constant of ELossFunction::UserPerObjMetric to the constants vector holding all the RegressionObjectives


camrongodbout:

andrey-khropov:
It is not a good idea because `UserPerObjMetric` can be used not only for regression but for other types of machine learning problems as well.

I've created an issue about this problem: https://github.com/catboost/catboost/issues/2936

PR #1543, opened 2021-01-19
Add functionality to restore R handle after deserializing

david-cortes:
This PR adds functionality to restore/complete/reconstruct an R model's internal C++ object after de-serializing an R model object, similar to XGBoost's `xgb.Booster.complete` function.

While this is done automatically when calling `catboost.predict` and similar, the model objects are not modified in-place, so the process needs to be repeated at each call, which makes it slow to use saved models that were serialized through R's typical functions. This function allows to avoid this restoration inside each function call.


kizill:
@arcadia-devtools Ship it!

arcadia-devtools:
CLA already signed

arcadia-devtools:
@kizill, internal review request created: [1600318](https://a.yandex-team.ru/review/1600318)

Issue #439, opened 2018-08-10
CRAN release?


cdeterman:
Problem: {No CRAN released version}

Is there a particular reason that there is no version of catboost released to the CRAN ecosystem for R?

annaveronika:
It would be great to make a CRAN release, but we need help with that.

cdeterman:
@annaveronika what help do you need?  I maintain multiple R packages and have helped others publish their packages.  I would be happy to be of assistance.

annaveronika:
https://tech.yandex.com/catboost/doc/dg/concepts/installation-docpage/ - here is the instruction, how to install the package.

We either provide the binaries or it is also possible to build from source, but we had many problems of people trying to build from source on windows, so the easiest way to install is to get the binaries.

We normally build using our own build system. But we also provide makefiles for python and cmdline. 
https://github.com/catboost/catboost/tree/master/make

I don't think we can upload the binary of the R package to CRAN.

Plus you need to have in mind that we are going to add GPU support to the package.

Given all this, what are the requirements and the first steps for catboost?

cdeterman:
Ideally we want to get the windows builds to work.  You are correct that binaries aren't allowed for submission to CRAN.  Strictly speaking Windows support isn't required but obviously something we should try to sort out.  Are there any dependencies that aren't normally installed on Windows or is the package self contained?  I tried installing it on my own Windows machine and ran in to a list of problems (truncated for space).


>Info: Attention! Using system user-defined compiler: None (check CC and CXX env vars).
Warn[-WNoInput]: in $S/contrib/libs/openssl/ya.make: macro for $S/contrib/libs/openssl/asm/windows/aes-x86_64.masm didn't use it as input
...
Error[-WNoOutput]: in $S/contrib/libs/openssl/ya.make: macro 9010:SRCSmasm=(x86_64cpuid.masm) resulted in no outputs, can't add to graph
...
Configure error (use -k to proceed)
make: *** [all] Error 1
Warning: running command 'make --no-print-directory -f "Makefile.win"' had status 2
ERROR: compilation failed for package 'catboost'



Is there something unique about your build system?

Does the package 'always' build successfully on Linux/OSX systems?  I don't see anything with respect to a continuous integration system that could be of use for this (e.g. Travis CI).

Regarding GPU support, that shouldn't be problem.  I am very familiar with this as I am the author of the `gpuR` package.  There are a few moving parts but it ultimately shouldn't be a major problem.

[...]

PR #1551, opened 2021-01-20
Update R docs

david-cortes:
This PR updates two particular topics about the R docs:
* It replaces usage of the `pool` object method `head` in the examples which looks like was tried to being used as a slicer or printer, but makes the examples fail due to the method not being able to work with non-numerical columns (hopefully another step towards https://github.com/catboost/catboost/issues/439). Along the way also fixes the example to use R's `factor` for categorical features instead of strings, which are not accepted by catboost.
* It adds missing prediction types to the docs which are accepted by the C++ functions but for some reason were not documented.


nikitxskv:
@arcadia-devtools Ship it!

arcadia-devtools:
CLA already signed

arcadia-devtools:
@nikitxskv, internal review request created: [1646069](https://a.yandex-team.ru/review/1646069)

PR #1555, opened 2021-01-21
(R) better class checks, avoid overflows

david-cortes:
This PR addresses two issues in the R code:
* The code tries to validate the class of objects by checking `class(obj) == "something"`, which works fine for running the code as-is but would not allow users  or other packages to sub-class objects from catboost. This changes it to the more idiomatic `inherits`.
* The prediction function might fail when it tries to reshape the output if the number of entries is larger than `INT_MAX` due to obtaining an integer from floating point division, which might not return the exact result. Such operation is not necessary however.


nikitxskv:
@arcadia-devtools Ship it!

arcadia-devtools:
CLA already signed

arcadia-devtools:
@nikitxskv, internal review request created: [1646067](https://a.yandex-team.ru/review/1646067)

PR #1556, opened 2021-01-21
More S3 methods for model class

david-cortes:
The R package has a class defined for catboost models, but this class is lacking two typical methods that one expects from model objects: `print` and `summary` - in particular, it's not nice to type e.g. `model` in the R console and get a long display of raw bytes which are uninformative.

This PR adds these methods for a nicer interactive experience.


catboost/R-package/R/catboost.R
```diff
@@ -396,6 +396,33 @@ print.catboost.Pool <- function(x, ...) {
...
 
+#' @title Print basic information about model
+#' @description Displays the most general characteristics of a CatBoost model.
+#' @param x The model obtained as the result of training.
+#' @param ... Not used
+#' @return The same model that was passed as input.
+#' @export
+print.catboost.Model <- function(x, ...) {
+    cat(sprintf("CatBoost model (%d trees)\n", x$tree_count))
+    cat(sprintf("Evaluation metric: %s\n", catboost.get_plain_params(x)$eval_metric))
```
nikitxskv, line 407: Let's replace "Evaluation metric" with "Loss function" and `$eval_metric` with `$loss_function`.

catboost/R-package/R/catboost.R
```diff
@@ -396,6 +396,33 @@ print.catboost.Pool <- function(x, ...) {
...
+    cat(sprintf("Fit to %d features\n", NROW(x$feature_importances)))
+    if (is.null.handle(x$handle))
+        cat("(Handle is incomplete)\n")
```
nikitxskv, line 410: What is this information for?

david-cortes: It’s in order to know if the model object doesn’t contain the underlying C++ object (happens after de-serializing through R’s own functions), so as to know if it’d benefit from calling the function to deserialize it from the raw bytes instead of having it internally de-serialize every time a function is called on the model.

catboost/R-package/R/catboost.R
david-cortes: Done.

nikitxskv:
@arcadia-devtools Ship it!

arcadia-devtools:
CLA already signed

arcadia-devtools:
@nikitxskv, internal review request created: [1666676](https://a.yandex-team.ru/review/1666676)

PR #1544, opened 2021-01-19
Retrieve R pointers outside loops

david-cortes:
The R interface tries to access the underlying data by retrieving pointers to the R's object data inside loops. This is not necessary as the pointer to the data can be obtained before entering a loop, saving operations by not requiring all the extra checks that happen before a pointer is retrieved.

Example:
```cpp
for (auto i=...)
    REAL(R_obj)[i] = ...
```
vs.
```cpp
double *ptr_R_obj = REAL(R_obj);
for (auto i=...)
    ptr_R_obj[i] = ...
```

**Disclaimer: I didn't manage to get the project's build system working, so I'm not 100% sure that this code will compile.**

I hereby agree to the terms of the CLA available at: https://yandex.ru/legal/cla/?lang=en

PR #1621, opened 2021-03-25
Fix retrieval of R pointers fixes #1620 ref #1544

### LINKED ISSUE #1620: INTEGER() can only be applied to a 'integer', not a 'NULL'
Problem:INTEGER() can only be applied to a 'integer', not a 'NULL'
catboost version:25.0 for R4.0.2
Operating System:Windows 8.1, Ubuntu 20.4
CPU:Intel
# GPU:NVIDIA

I get in both Windows and Ubuntu R

Error in catboost.from_matrix(as.matrix(preprocessed), label, cat_features,  : 
  INTEGER() can only be applied to a 'integer', not a 'NULL'

when I execute the demo code

library(catboost)

train_path = system.file("extdata", 
                         "adult_train.1000", 
                         package="catboost")
test_path = system.file("extdata", 
                        "adult_test.1000", 
                        package="catboost")

column_description_vector = rep('numeric', 15)
cat_features <- c(3, 5, 7, 8, 9, 10, 11, 15)
for (i in cat_features)
  column_description_vector[i] <- 'factor'

train <- read.table(train_path, 
                    head = F, 
                    sep = "\t", 
                    colClasses = column_description_vector, 
                    na.strings='NAN')
test <- read.table(test_path, 
                   head = F, 
                   sep = "\t", 
                   colClasses = column_description_vector, 
                   na.strings='NAN')
target <- c(1)
train_pool <- catboost.load_pool(data=train[,-target], 
                                 label = train[,target])

david-cortes:
This PR fixes an issue introduced by previous PR https://github.com/catboost/catboost/pull/1544 (reported in https://github.com/catboost/catboost/issues/1620), in which many R functions would fail due to trying to get a pointer from an object of class `NULL`, by making a quick check in the C++ code (an alternative fix would be to convert those NULLs to empty vectors inside R).

Disclaimer: I am not able to run the build system and am not 100% sure that this will compile, neither am I able to test that the R functions will run correctly afterwards. Would be helpful if someone could rebuild it and check that the example in the issue runs correctly, or that it doesn't introduce new failing tests from the R package's own tests.

**Would be ideal if this could be incorporated into a quick fix in the release version, as version 0.25 is otherwise pretty unusable in R.**

### DIFF
--- a/catboost/R-package/src/catboostr.cpp
+++ b/catboost/R-package/src/catboostr.cpp
@@ -230,7 +230,7 @@ EXPORT_FUNCTION CatBoostCreateFromMatrix_R(SEXP matrixParam,
 
         visitor->Start(metaInfo, dataRows, EObjectsOrder::Undefined, {});
 
-        double *ptr_targetParam = REAL(targetParam);
+        double *ptr_targetParam = Rf_isNull(targetParam)? nullptr : REAL(targetParam);
         for (auto targetIdx : xrange(targetColumns)) {
             TVector<float> target(targetRows);
             for (auto docIdx : xrange(targetRows)) {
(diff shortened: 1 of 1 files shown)


kizill:
@arcadia-devtools Ship it!

kizill:
Thanks a lot, strange that internal tests don'catch a problem, will figure out why.

david-cortes:
> Thanks a lot, strange that internal tests don'catch a problem, will figure out why.

Thing is, the R tests do catch the problem, but many tests have been failing since many versions ago. The R tests can be run with `R CMD check <file>.tar.gz`, but the result saying they failed is only in stdout.

arcadia-devtools:
CLA already signed

arcadia-devtools:
@kizill, internal review request created: [1709694](https://a.yandex-team.ru/review/1709694)

arcadia-devtools:
Check [1709694](https://a.yandex-team.ru/review/1709694) passed

PR #1668, opened 2021-04-23
(R) Update C++ handles by reference to avoid redundant copies

david-cortes:
This PR changes the structure used behind the pointers to C++ objects, from an R list (copy-on-write) to an R environment (pass-by-reference).

Currently, as it uses a list with copy-on-modify semantics, when a model object has a reference to a null pointer (happens after restarting an R session or de-serializing an object through R's own functions) and the C++ object is reconstructed from the serialized bytes, it will reconstruct the object in a local environment, which makes it short lived as it doesn't update the original object outside of the function being called, having to reconstruct the object at each successive function call (which is slow and causes extra memory usage).

After the changes, it will only need to de-serialize the C++ object once, as the original object will have a pointer that is updated by reference. It also avoids potential segfaults if the user calls a function that modifies the C++ object but does not re-assign the result to the model object holder from outside.

The changes also allow shortening a bit the R source code as it can call a single function to check the model class and validity of the pointer instead of having to copy-paste the same code over and over.

### DIFF
--- a/catboost/R-package/R/catboost.R
+++ b/catboost/R-package/R/catboost.R
@@ -415,7 +415,7 @@ print.catboost.Model <- function(x, ...) {
     cat(sprintf("CatBoost model (%d trees)\n", x$tree_count))
     cat(sprintf("Loss function: %s\n", catboost.get_plain_params(x)$loss_function))
     cat(sprintf("Fit to %d features\n", NROW(x$feature_importances)))
-    if (is.null.handle(x$handle))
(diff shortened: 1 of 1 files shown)


kizill:
Wow, that's a really great speedup for large models, thank you!

kizill:
@arcadia-devtools Ship it!

arcadia-devtools:
CLA already signed

arcadia-devtools:
@kizill, internal review request created: [1776889](https://a.yandex-team.ru/review/1776889)

PR #1672, opened 2021-04-23
Issue #252 (eval_metrics in R)

### LINKED ISSUE #252: eval_metrics in R
The python library has eval_metrics method, which calculates metric value on every iteration and returns for each requested metric its values on each iteration. We need to do the same for R. This will also allow for visualisation of metric values.

ebalukova:
Resolves #252
I hereby agree to the terms of the CLA available at: [link](https://yandex.ru/legal/cla/).


catboost/R-package/src/catboostr.cpp
```diff
@@ -958,4 +962,97 @@ EXPORT_FUNCTION CatBoostEvaluateObjectImportances_R(
...
+        }
+    }
+
+    TVector<TVector<double>> metricsScore = plotCalcer.GetMetricsScore();
+    plotCalcer.SaveResult(CHAR(asChar(resultDirParam)), /*metricsFile=*/"", /*saveMetrics*/ false, /*saveStats=*/true).ClearTempFiles();
+
+    auto metricsResult = CreateMetricsFromDescription(metricDescriptions, model->GetDimensionsCount());
+    metricsCount = metricsResult.ysize();
+
+    result = PROTECT(allocVector(VECSXP, metricsCount));
```
david-cortes, line 1036: I think if this allocation or the next ones fail it will cause a memory leak due to not destructing the C++ objects from the earlier code. Is the size known in advance from the input parameters? Could it be allocated earlier?

grag90 (approved):

kizill:
@arcadia-devtools Ship it!

arcadia-devtools:
CLA already signed

arcadia-devtools:
@kizill, internal review request created: [1776888](https://a.yandex-team.ru/review/1776888)

PR #1680, opened 2021-04-28
Add support for multitask regression with missing values on targets

gmrandazzo:
With this patch I would like to propose the possibility to run multitask regression with missing values on tasks.
Since the current catboost version do not accept "nan" values on targets, I have defined as "missing value" the value -9999. 
This value is checked and skiped inside the loss function using the FuzzyEqual() function. When -9999 is found on targets, it will automatically skipped and not considered in the RMSE calculation. 

To use this function you must specify loss="MultiRMSEWithMissingValues" and -9999 value on the task were the value is "missing".

This piece of code has been compiled and tested. 

There is also a test in pytest which is a copy of the "multiregression test" plus 10 missing values on task 1 and task 2. 
The test seems to work fine. 

This work was supported by [F. Hoffmann-La Roche Ltd](https://www.roche.com/) through a Postdoc Fellowship Programme

Many thanks for your consideration.
Best regards
GMR

1. Read [instructions for contributors](https://catboost.ai/docs/concepts/development-and-contributions.html). -> Done
2. Run `ya make` in catboost folder to make sure the code builds. -> Code builds and it works
3. Add tests that test your change. -> Done test has been added 
4. Run tests using `ya make -t -A` command. -> Test working
5. If you haven't already, complete the [CLA](https://yandex.ru/legal/cla/). -> I hereby agree to the terms of the CLA available at: https://yandex.ru/legal/cla/?lang=en


LyzhinIvan:
Hi, @gmrandazzo! 

Why do you think that missing values in target require new loss function? Why not just delete these objects from the dataset?

gmrandazzo:
> Hi, @gmrandazzo!
> 
> Why do you think that missing values in target require new loss function? Why not just delete these objects from the dataset?

Hi @LyzhinIvan 

the question is that we normally want to exploit all the information contained on our tasks without removing instances from our dataset. Our main principle is that "similar instances may have similar target properties" and the similarity is defined in the feature space. We believe that there is information in tasks that could be relevant to other tasks where the information is missing.

LyzhinIvan:
1) CatBoost source code contains many places that expect target existence. For example, processing of categorical, text, embedding features. Your approach could work only for float features (but I'm not sure). So it is required to add a couple of asserts at least. 
2) Your approach assumes that we should add a clone for each loss function in the future (LoglossWithMissing, MAEWithMissing, YetiRankWIthMissing, etc). There should be more elegant solution.
3) Actually, I see only one place where objects with missing values could have any effect - pool quantization (to calculate borders taking into account all objects). And there is such workaround - quantize all features without target, save quantization borders, quantize features with non-missed target  using saved quantization borders. Sample code:
```
all_features = Pool(X_full)
all_features.quantize()
all_features.save_quantization_borders('borders.txt')
pool = Pool(X_clean, y_clean)
pool.quantize(input_borders='borders.txt')
```

gmrandazzo:
> 1. CatBoost source code contains many places that expect target existence. For example, processing of categorical, text, embedding features. Your approach could work only for float features (but I'm not sure). So it is required to add a couple of asserts at least.

Yes. The code works for "floating points" features and targets. I have not tested with "categorical" or "text" ones. I can give a try and make it "friendly" 

> 2. Your approach assumes that we should add a clone for each loss function in the future (LoglossWithMissing, MAEWithMissing, YetiRankWIthMissing, etc). There should be more elegant solution.

Yes... That's a proposal. A more elegant way could be for example the integration of "missing value" check on all the current existent loss functions. If you agree, will give a try to it and propose a new patch.

> 3. Actually, I see only one place where objects with missing values could have any effect - pool quantization (to calculate borders taking into account all objects). And there is such workaround - quantize all features without target, save quantization borders, quantize features with non-missed target  using saved quantization borders. Sample code:
> 
> ```
> all_features = Pool(X_full)
> all_features.quantize()
> all_features.save_quantization_borders('borders.txt')
> pool = Pool(X_clean, y_clean)
> pool.quantize(input_borders='borders.txt')
> ```

Thank you for the suggestion. Would be interesting to see how the "pool quantization" performs in comparison with my current integration.

gmrandazzo:
Hi Ivan,

Here a new proposal.

I tried to rework in a more "elegant" way the support for missing values (always defined with the -9999 value) and directly handled with the current available "MultiRMSE" loss function instead of having a clone MultiRMSEWithtMissingValues. 

Than I also noticed that on catboost/catboost/libs/metrics/optimal_const_for_loss.h, starting from 0 is better than starting from the average of target when missing values are available. The algorithm converge well to low RMSE values and also produce better results. 

Conversely, if no missing values are present on targets, starting from 0 or starting from the average of targets shows similar results.

For that I have introduced a function on optimal_const_for_loss.h  that catch if there are missing value on targets and return "Nothing();" or return the average of targets if no missing values are found.

Bests

gmrandazzo:
Dear @LyzhinIvan and catboost mantainers :)
 
after a nice personal chat with @LyzhinIvan (many thanks for your support), here I come up with a new clean and more elegant proposal for the inclusion of multitask regression with missing values on target and with this message I would like to comment those changes.

To give a bit of context on the topic, normally missing values occurs on multitask regression when instances/objects may have values in one task but not in others. This is usually the case where measurements/properties are not acquired in all tasks/targets for several reasons and we want to leverage all the available instance/objects space without removing the incomplete tasks. With this proposal, we would like to introduce the possibility to run multitask learning experiments that consider those missing values and offer mainly 2 principal advantages in catboost:

- improve the data efficiency avoiding to discard instances with missing values
- leveraging the auxiliary information contained in linked/unlinked tasks to prevent overfitting and produce better predictions.

Here comes the explanation for the code modification. In general missing values are defined with "nan" values. In order to integrate those ones on targets:

-  I have first modified target.cpp by substituting “CB_ESURE(...” to skip nan with a function that converts a nan value into a MISSING_VALUE = -9999, which is defined in ymath.h. Not sure if this MISSING_VALUE definition is in the right place. However it can be easy moved into another more compliant place. This conversion of nan -> MISSING_VALUE avoid problems  BuildTargetClassifier->SelectBorders->BestSplit methods. 

- As you may imagine, the introduction of the conversion of nan into -9999, will pose a problem for single task loss functions. Indeed, -9999 is a float value that is considered for the calculation of the first and second order derivatives. Then to prevent the use of MISSING_VALUE on single task regression problems I have introduced on data_providers.cpp an assertion that check for single task regression loss function and that check if "nan" values are present on target. If yes this assertion will throw an error preventing thus to use with single task regression models the nan and then the -9999 on target. To prevent this “check for nan” on targets I have then added a method on “TrawTargetDataProvider” named TargetsContainsNan() which check on “float” targets if “nan” are present.


- To check for a missing value now there will be a function defined on catboost/private/libs/options/enum_helpers.h which check if a target is a missing value “IsTargetMissingValue(float value)”. Not sure if this is the right place. However could be moved into another more compliant place.

- Again, MultiRMSE now will automatically consider this “MISSING_VALUE” definition and skip on the whole first and second derivative calculation.

- The problem of “initialization with 0” instead of the initialization with the average of targets is solved. It was a bug of the previous code. Thanks to Ivan which pointed me in the right direction.


- Command line tests have been updated accordingly with the current code proposal version and looks good.

Bests
GMR

ThomasWolf0701:
> 
> 
> Dear @LyzhinIvan and catboost mantainers :)
> 
> after a nice personal chat with @LyzhinIvan (many thanks for your support), here I come up with a new clean and more elegant proposal for the inclusion of multitask regression with missing values on target and with this message I would like to comment those changes.
> 
> To give a bit of context on the topic, normally missing values occurs on multitask regression when instances/objects may have values in one task but not in others. This is usually the case where measurements/properties are not acquired in all tasks/targets for several reasons and we want to leverage all the available instance/objects space without removing the incomplete tasks. With this proposal, we would like to introduce the possibility to run multitask learning experiments that consider those missing values and offer mainly 2 principal advantages in catboost:
> 
>     * improve the data efficiency avoiding to discard instances with missing values
> 
>     * leveraging the auxiliary information contained in linked/unlinked tasks to prevent overfitting and produce better predictions.
> 
> 
> Here comes the explanation for the code modification. In general missing values are defined with "nan" values. In order to integrate those ones on targets:
> 
>     * I have first modified target.cpp by substituting “CB_ESURE(...” to skip nan with a function that converts a nan value into a MISSING_VALUE = -9999, which is defined in ymath.h. Not sure if this MISSING_VALUE definition is in the right place. However it can be easy moved into another more compliant place. This conversion of nan -> MISSING_VALUE avoid problems  BuildTargetClassifier->SelectBorders->BestSplit methods.
> 
>     * As you may imagine, the introduction of the conversion of nan into -9999, will pose a problem for single task loss functions. Indeed, -9999 is a float value that is considered for the calculation of the first and second order derivatives. Then to prevent the use of MISSING_VALUE on single task regression problems I have introduced on data_providers.cpp an assertion that check for single task regression loss function and that check if "nan" values are present on target. If yes this assertion will throw an error preventing thus to use with single task regression models the nan and then the -9999 on target. To prevent this “check for nan” on targets I have then added a method on “TrawTargetDataProvider” named TargetsContainsNan() which check on “float” targets if “nan” are present.
> 
>     * To check for a missing value now there will be a function defined on catboost/private/libs/options/enum_helpers.h which check if a target is a missing value “IsTargetMissingValue(float value)”. Not sure if this is the right place. However could be moved into another more compliant place.
> 
>     * Again, MultiRMSE now will automatically consider this “MISSING_VALUE” definition and skip on the whole first and second derivative calculation.
> 
>     * The problem of “initialization with 0” instead of the initialization with the average of targets is solved. It was a bug of the previous code. Thanks to Ivan which pointed me in the right direction.
> 
>     * Command line tests have been updated accordingly with the current code proposal version and looks good.
> 
> 
> Bests
> GMR

If I understand this correctly missing values would now be defined as nan instead of as -9999. While this sounds reasonable, I think scikit-learn might throw an error if it detects nan in the target, while with a number it should work and could just write a score function that deals with this as needed. Am not sure if this works with nan. Would it be possible to have some user defined way to define what is a missing value ? Many thanks for your efforts.

gmrandazzo:
Hi Thomas,

well, yes.  So now missing values are "nan".
Indeed, forcing user to change their missing values (nan) into -9999 was obliviously not the right way.

To be more clear, with this current patch update the "nan" values are automatically handled and converted inside the CatBoost into -9999. Of course there is also a reason of why is better -9999 and not another value such "0". This is because normally in our real world, "0" in regression means something real. Instead is very difficult to find something that is real with a -9999 value. 

Finally, the current patch introduce an error when "Single task" regression finds "nan" on a target. However allows to use "nan" on targets and to consider this as missing values only on multitask regression case. 

Then, to answer your question in regards to the user defined missing value, nan in my opinion is the best way to do.
If you look indeed at Pandas, if you have  a csv with missing values, they will be automatically converted into "nan" [Please have a look here](https://pandas.pydata.org/pandas-docs/stable/user_guide/missing_data.html). 

Here also an example:

```
cat example.csv
1,,3,4
,6,7,
9,10,,12

python
>>> import pandas as pd
>>> d = pd.read_csv("example.csv", header=None)
>>> d.values
array([[ 1., nan,  3.,  4.],
       [nan,  6.,  7., nan],
       [ 9., 10., nan, 12.]])
```

My fear is that converting this into an user defined values became too "flexible" and will probably make confusions and problems for the loss functions. You then may have problems with single task regression cases and so on. So fixing and considering "nan" to define missing values I guess is a good compromise.

Moreover in CatBoost there is support for several Nan definitions ( Please have a look to [loader.cpp](https://github.com/catboost/catboost/blob/master/catboost/libs/data/loader.cpp#L404)) and the current patch update use this internal nan handling which I found to be a good choice. 

Thank you for the interest and proposal 👍🏻

gmrandazzo:
Dear all,

Here a new patch proposition.
 
After another round discussion with  @LyzhinIvan, we thought that the conversion of nan into -9999 is a bad idea. Here some reasons. Even if it is a "quick" way to propose a patch, this way could be error prone especially on the calculation of "Borders". Moreover with this solution we can not be sure that everything works correctly.  That's why instead of adopting the -9999 conversion now with this patch version we propose to adopt directly "nan"  as missing values. To achieve that, I propose a modification on "target_classifier.cpp" to include the possibility to skip for "nan" while calculating the Borders on binarization.cpp. For that, on binarization.cpp, now if "featureValuesMayContainNans" is True will automatically "remove" all the "nan"  values before calculating borders and if False and a "nan" value is found will throw an error. 

We also agreed that would be a good solution to "add" a new metrics and error function instead of modifying the already available MultiRMSE for 2 reasons:

1) speed effect on MultiRMSE
2) the weight sum on the metric should be calculated individually in each dimension skipping the nan values. The gradient and the hessian should be set to 0 for missing values.

For that reason, I propose also this loss function name "MultiRMSEWithMissingValues" which calculates the error according the formula:

formula^2w_{i}1_{t_{i,d\neq%20nan}}}{\sum_{N}^{i=1}w_{i}1_{t_{i,d\neq%20nan}}})

Thank you again for your support. 

Bests,
GMR

catboost/libs/metrics/metric.cpp
```diff
@@ -466,6 +467,74 @@ void TMultiRMSEMetric::GetBestValue(EMetricBestValue* valueType, float* /*bestVa
...
+    const auto evalImpl = [=](bool useWeights, bool hasDelta) {
+        const auto realApprox = [=](int dim, int idx) { return approx[dim][idx] + (hasDelta ? approxDelta[dim][idx] : 0); };
+        const auto realWeight = [=](int idx) { return useWeights ? weight[idx] : 1; };
+
+        TMetricHolder error(1);
+        for (auto dim : xrange(target.size())) {
+            double w = 0.f;
+            double e = 0.f;
+            for (auto i : xrange(begin, end)) {
+                if (std::isnan((float)target[dim][i])) {
```
LyzhinIvan, line 514: let's inverse `if` condition

catboost/libs/metrics/metric.cpp
```diff
@@ -466,6 +467,74 @@ void TMultiRMSEMetric::GetBestValue(EMetricBestValue* valueType, float* /*bestVa
...
+    int begin,
+    int end
+) const {
+    const auto evalImpl = [=](bool useWeights, bool hasDelta) {
+        const auto realApprox = [=](int dim, int idx) { return approx[dim][idx] + (hasDelta ? approxDelta[dim][idx] : 0); };
+        const auto realWeight = [=](int idx) { return useWeights ? weight[idx] : 1; };
+
+        TMetricHolder error(1);
+        for (auto dim : xrange(target.size())) {
+            double w = 0.f;
```
LyzhinIvan, line 511:
```
double sumWeights = 0.0;
double sumErrors = 0.0;
```

catboost/libs/metrics/metric.cpp
```diff
@@ -466,6 +467,74 @@ void TMultiRMSEMetric::GetBestValue(EMetricBestValue* valueType, float* /*bestVa
...
+            double e = 0.f;
+            for (auto i : xrange(begin, end)) {
+                if (std::isnan((float)target[dim][i])) {
+                    continue;
+                } else {
+                    e += realWeight(i) * Sqr(realApprox(dim, i) - target[dim][i]);
+                    w += realWeight(i);
+                }
+            }
+            error.Stats[0] += e/w;
```
LyzhinIvan, line 521: It isn't correct because `EvalSingleThread` can be called several times for batches of dataset and then `GetFinalError` will be called. You should add `sumErrors` to `Stats[dim*2]` and `sumWeights` to `Stats[dim*2+1]`. Then in the `GetFinalError` you should divide `Stats[dim*2]` by `Stats[dim*2+1]` for each dimension and sum all results.

catboost/private/libs/target/data_providers.cpp
```diff
@@ -529,6 +537,12 @@ namespace NCB {
                 IsMultiRegressionObjective(mainLossFunction->GetLossFunction()) || rawData.GetTargetDimension() <= 1,
                 "Currently only multi-regression objectives work with multidimensional target"
             );
+            
+            if(IsRegressionObjective(mainLossFunction->GetLossFunction()))
+                CB_ENSURE(rawData.TargetsContainsNan() == false,
```
LyzhinIvan, line 542: `!rawData.TargetsContainsNan()`

catboost/private/libs/target/target_converter.cpp
```diff
@@ -50,8 +49,11 @@ namespace NCB {
             size_t i = 0;
             (*floatSequence)->ForEach(
                 [resultRef, &i] (float value) {
-                    CB_ENSURE(!std::isnan(value), "NaN values are not supported for target");
-                    resultRef[i++] = value;
+                    if (std::isnan(value)) {
```
LyzhinIvan, line 52: Why not just `resultRef[i++] = value;`?

LyzhinIvan (changes requested):

catboost/private/libs/algo_helpers/error_functions.h
```diff
@@ -216,6 +216,43 @@ class TMultiRMSEError final : public TMultiDerCalcer {
...
+        TConstArrayRef<double> approx,
+        TConstArrayRef<float> target,
+        float weight,
+        TVector<double>* der,
+        THessianInfo* der2
+    ) const override {
+        const int dim = target.size();
+        for (auto i : xrange(dim)) {
+            if (std::isnan((float)target[i])) {
+                (*der)[i] = 0.f;
```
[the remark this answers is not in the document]
gmrandazzo: Fixed.

catboost/libs/data/target.cpp
```diff
@@ -417,6 +417,30 @@ ERawTargetType TRawTargetDataProvider::GetTargetType() const {
...
+    if(!Data.Target.empty() && GetTargetType() == ERawTargetType::Float) {
+        for (auto targetIdx : xrange(GetTargetDimension())) {
+            TVector<float> targetNumeric(GetObjectCount());
+            TArrayRef<float> targetNumericRef = targetNumeric;
+            ToArray(*Get<ITypedSequencePtr<float>>(Data.Target[targetIdx]), targetNumericRef);
+            for (auto objectIdx : xrange(targetNumericRef.size())) {
+                if(std::isnan(targetNumericRef[objectIdx])){
+                    return true;
+                }
+                else{
```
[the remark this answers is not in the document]
gmrandazzo: Fixed.

catboost/private/libs/algo/target_classifier.cpp
```diff
@@ -50,6 +51,7 @@ TTargetClassifier BuildTargetClassifier(
 
     CB_ENSURE(!target.empty(), "train target should not be empty");
 
+    bool featureValuesMayContainNans = false;
```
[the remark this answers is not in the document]
gmrandazzo: Fixed.

catboost/private/libs/target/target_converter.cpp
gmrandazzo: Fixed.

catboost/libs/metrics/optimal_const_for_loss.h
```diff
@@ -16,16 +16,28 @@
...
+                } else {
+                    targetSum += target[i];
+                    summaryWeight += 1.0;
+                }
             }
-        }
+         } else {
+             Y_ASSERT(target.size() == weights.size());
+            for (size_t i = 0; i < target.size(); ++i){
+                if (std::isnan((float)target[i])) {
```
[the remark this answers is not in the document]
gmrandazzo: Fixed.

catboost/libs/metrics/metric.cpp
gmrandazzo: Fixed.

catboost/libs/metrics/metric.cpp
gmrandazzo: Fixed.

catboost/libs/metrics/metric.cpp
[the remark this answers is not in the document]
gmrandazzo: Fixed.

catboost/private/libs/target/data_providers.cpp
gmrandazzo: Fixed.

gmrandazzo (commented):
Changes approved and implemented in the current code version.
Thank you.

catboost/libs/data/target.cpp
```diff
@@ -417,6 +417,25 @@ ERawTargetType TRawTargetDataProvider::GetTargetType() const {
...
+            TVector<float> targetNumeric(GetObjectCount());
+            TArrayRef<float> targetNumericRef = targetNumeric;
+            ToArray(*Get<ITypedSequencePtr<float>>(Data.Target[targetIdx]), targetNumericRef);
+            for (auto objectIdx : xrange(targetNumericRef.size())) {
+                if(std::isnan(targetNumericRef[objectIdx])){
+                    return true;
+                }
+            }
+        }
+        return false;
```
LyzhinIvan, line 433: redundant

catboost/libs/data/target.h
```diff
@@ -135,6 +135,8 @@ namespace NCB {
 
         ERawTargetType GetTargetType() const;
 
+        bool TargetsContainsNan() const;
```
LyzhinIvan, line 138: IsTargetsContainsNan

catboost/libs/metrics/metric.cpp
```diff
@@ -466,6 +467,79 @@ void TMultiRMSEMetric::GetBestValue(EMetricBestValue* valueType, float* /*bestVa
...
+        return error;
+    };
+
+    return DispatchGenericLambda(evalImpl, !weight.empty(), !approxDelta.empty());
+}
+
+double TMultiRMSEWithMissingValues::GetFinalError(const TMetricHolder& error) const {
+    double finalError = 0.0;
+    for (size_t dim = 0; dim < error.Stats.size(); dim+=2) {
+        finalError += error.Stats[dim]/error.Stats[dim+1];
```
LyzhinIvan, line 534: We should check `error.Stats[dim+1]` for 0

catboost/libs/metrics/metric.cpp
```diff
@@ -466,6 +467,79 @@ void TMultiRMSEMetric::GetBestValue(EMetricBestValue* valueType, float* /*bestVa
...
+        TMetricHolder error(target.size()*2);
+        for (auto dim : xrange(target.size())) {
+            double sumWeights = 0.0;
+            double sumErrors = 0.0;
+            for (auto i : xrange(begin, end)) {
+                if (!std::isnan(target[dim][i])) {
+                    sumErrors += realWeight(i) * Sqr(realApprox(dim, i) - target[dim][i]);
+                    sumWeights += realWeight(i);
+                } else {
+                    continue;
```
LyzhinIvan, line 518: redundant

catboost/libs/metrics/optimal_const_for_loss.h
```diff
@@ -16,16 +16,28 @@
...
-        } else {
-            Y_ASSERT(target.size() == weights.size());
+         if (weights.empty()) {
             for (size_t i = 0; i < target.size(); ++i) {
-                targetSum += target[i] * weights[i];
+                 if (!std::isnan(target[i])) {
+                    targetSum += target[i];
+                    summaryWeight += 1.0;
+                } else {
+                    continue;
```
LyzhinIvan, line 27: redundant

catboost/libs/metrics/optimal_const_for_loss.h
```diff
@@ -16,16 +16,28 @@
...
             }
-        }
+         } else {
+             Y_ASSERT(target.size() == weights.size());
+            for (size_t i = 0; i < target.size(); ++i){
+                if (!std::isnan(target[i])) {
+                    targetSum += target[i] * weights[i];
+                    summaryWeight += weights[i];
+                } else {
+                    continue;
```
LyzhinIvan, line 37: redundant

catboost/private/libs/algo/target_classifier.cpp
```diff
@@ -19,11 +19,12 @@ static TVector<float> SelectBorders(
     TConstArrayRef<float> target,
     int targetBorderCount,
     EBorderSelectionType targetBorderType,
-    bool allowConstLabel) {
+    bool allowConstLabel,
+    bool featureValuesMayContainNans) {
```
LyzhinIvan, line 23: `featureValuesMayContainNans` -> `targetValuesMayContainNans`

catboost/private/libs/algo/target_classifier.cpp
```diff
@@ -59,7 +61,12 @@ TTargetClassifier BuildTargetClassifier(
         case ELossFunction::RMSE:
         case ELossFunction::MultiRMSE:
             return TTargetClassifier(
-                SelectBorders(target, targetBorderCount, targetBorderType, allowConstLabel),
+                SelectBorders(target, targetBorderCount, targetBorderType, allowConstLabel, targetValuesMayContainNans),
+                targetId);
+        case ELossFunction::MultiRMSEWithMissingValues:
+            targetValuesMayContainNans = true;
+            return TTargetClassifier(
+                SelectBorders(target, targetBorderCount, targetBorderType, allowConstLabel, targetValuesMayContainNans),
```
LyzhinIvan, line 69: Let's set default value for `targetValuesMayContainNans` in `SelectBorders` to `false` and will pass `true` only here.

catboost/private/libs/algo_helpers/error_functions.h
```diff
@@ -216,6 +216,43 @@ class TMultiRMSEError final : public TMultiDerCalcer {
...
+    void CalcDers(
+        TConstArrayRef<double> approx,
+        TConstArrayRef<float> target,
+        float weight,
+        TVector<double>* der,
+        THessianInfo* der2
+    ) const override {
+        const int dim = target.size();
+        for (auto i : xrange(dim)) {
+            if (std::isnan((float)target[i])) {
```
LyzhinIvan, line 234: conversion to float is redundant, `target[i]` is already `float`

catboost/private/libs/algo_helpers/error_functions.h
```diff
@@ -216,6 +216,43 @@ class TMultiRMSEError final : public TMultiDerCalcer {
...
+            }
+        }
+
+        if (der2 != nullptr) {
+            Y_ASSERT(der2->HessianType == EHessianType::Diagonal &&
+                     der2->ApproxDimension == dim);
+
+            for (auto i : xrange(dim)) {
+                if (std::isnan((float)target[i])) {
+                    der2->Data[i] = 0.f;
```
LyzhinIvan, line 247: `0.0`

catboost/private/libs/algo_helpers/error_functions.h
LyzhinIvan, line 234: `const float realWeight = weight * std::isnan(target[i]);` allows to get rid of `if` statements

catboost/private/libs/options/enum_helpers.cpp
```diff
@@ -12,6 +12,8 @@
 #include <util/generic/map.h>
 #include <util/generic/ptr.h>
 #include <util/string/cast.h>
+#include <util/generic/ymath.h>
```
LyzhinIvan, line 15: Why it is required here?

library/cpp/grid_creator/binarization.cpp
```diff
@@ -86,12 +86,17 @@ namespace NSplitSelection {
...
         }
-
-        auto firstNanPos = std::remove_if(features.Values.begin(), features.Values.end(), IsNan);
-        if (firstNanPos != features.Values.end()) {
-            if (featureValuesMayContainNans) {
+        
+        if(featureValuesMayContainNans){
+            // Remove all nan to calculate borders in MultiRMSEWithMissingValues
+            auto firstNanPos = std::remove_if(features.Values.begin(), features.Values.end(), IsNan);
+            while(firstNanPos != features.Values.end()){
```
LyzhinIvan, line 93: What the purpose of `while` here? Sorry, I can't understand why this block of code has been changed.

LyzhinIvan (changes requested):

catboost/libs/data/target.cpp
gmrandazzo: Fixed.

catboost/libs/data/target.h
gmrandazzo: Fixed.

catboost/libs/metrics/metric.cpp
gmrandazzo: Fixed

catboost/libs/metrics/metric.cpp
gmrandazzo: Fixed

catboost/libs/metrics/optimal_const_for_loss.h
gmrandazzo: Fixed.

catboost/private/libs/algo_helpers/error_functions.h
gmrandazzo: Fixed

catboost/private/libs/algo_helpers/error_functions.h
gmrandazzo: Fixed.

catboost/private/libs/options/enum_helpers.cpp
gmrandazzo: Fixed. It was present from an old code proposal.

catboost/private/libs/algo_helpers/error_functions.h
gmrandazzo: Awesome trick! :) Fixed.

library/cpp/grid_creator/binarization.cpp
gmrandazzo: Ok. This while is necessary because all the "nan" value on a target needs to be "removed" before the border calculation using the BestSplit method. The old version was checking for only one nan to remove. 

Whit this value, if is a multiregression with missing value then there is a while loop that check and remove all nan before calculating the borders. Else will stop. Is this answering to your question?
Many thanks 👍🏻

catboost/libs/metrics/optimal_const_for_loss.h
gmrandazzo: Fixed.

catboost/private/libs/algo/target_classifier.cpp
gmrandazzo: Fixed.

catboost/private/libs/algo/target_classifier.cpp
gmrandazzo: Fixed.

library/cpp/grid_creator/binarization.cpp
LyzhinIvan: The "old" version removes all NaNs. `remove_if` moves all NaNs at the end of vector and returns the first position of NaN (https://en.cppreference.com/w/cpp/algorithm/remove). `erase` deletes this obtained tail of NaNs (https://en.cppreference.com/w/cpp/container/vector/erase).

gmrandazzo: You are right and I see the point. Tested and there is no need of "while" here. Code reverted back. Thank you.

catboost/private/libs/algo_helpers/error_functions.h
LyzhinIvan: Ahh, yes, you are right.

gmrandazzo (commented):
done.

catboost/private/libs/options/catboost_options.cpp
```diff
@@ -667,7 +673,7 @@ void NCatboostOptions::TCatBoostOptions::Validate() const {
         // we may adjust non-set BoostFromAverage in data dependant tuning
         CB_ENSURE(EqualToOneOf(lossFunction, ELossFunction::RMSE, ELossFunction::Logloss,
             ELossFunction::CrossEntropy, ELossFunction::Quantile, ELossFunction::MAE, ELossFunction::MAPE,
-            ELossFunction::MultiRMSE),
+            ELossFunction::MultiRMSE, ELossFunction::MultiRMSEWithMissingValues),
             "You can use boost_from_average only for these loss functions now: " <<
             "RMSE, Logloss, CrossEntropy, Quantile, MAE, MAPE or MultiRMSE.");
```
LyzhinIvan, line 678: add MultiRMSEWithMissingValues to message

catboost/libs/metrics/optimal_const_for_loss.h
```diff
@@ -16,16 +16,24 @@
 
 namespace NCB {
     inline float CalculateWeightedTargetAverage(TConstArrayRef<float> target, TConstArrayRef<float> weights) {
```
LyzhinIvan, line 18: `boost_from_average` does not work for MultiRMSEWithMissingValues now - it is required to add it near 151 line in this file

catboost/private/libs/target/data_providers.cpp
```diff
@@ -494,8 +494,16 @@ namespace NCB {
                 << " loss/metrics require target data"
             );
         }
-
-        const bool needCheckTarget = !mainLossFunction || !IsRegressionObjective(mainLossFunction->GetLossFunction());
+        bool needCheckTarget;
+        if(mainLossFunction->GetLossFunction() == ELossFunction::MultiRMSEWithMissingValues){
```
LyzhinIvan, line 498: missing spaces `if (` and `) {`

catboost/libs/metrics/metric.cpp
```diff
@@ -544,6 +545,84 @@ TVector<TParamSet> TMultiRMSEMetric::ValidParamSets() {
     return {TParamSet{{TParamInfo{"use_weights", false, true}}, ""}};
 };
 
+/* MultiRMSEWithMissingValues */
```
LyzhinIvan, line 548: let's add check for NaNs for all metrics in CheckPreprocessedTarget https://github.com/catboost/catboost/blob/74c9ca7e62d01236199e1bf44decd05586df6624/catboost/libs/metrics/metric.cpp#L6447 for fast fail if user tries to calc something with NaNs except of MultiRMSEWithMissingValues and UserDefined metrics

catboost/private/libs/target/data_providers.cpp
```diff
@@ -531,6 +539,12 @@ namespace NCB {
                 rawData.GetTargetDimension() <= 1,
                 "Currently only multi-regression and survival objectives work with multidimensional target"
             );
+
```
LyzhinIvan, line 542: I think this check will be redundant after adding of checks to `CheckPreprocessedTarget`

LyzhinIvan (changes requested):

library/cpp/grid_creator/binarization.cpp
```diff
@@ -86,7 +86,7 @@ namespace NSplitSelection {
                 throw (yexception() << "Unexpected Nan value.");
             }
         }
-
+
```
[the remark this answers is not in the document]
gmrandazzo: Fixed

catboost/pytest/data/multiregression_with_missing/test.txt
```diff
@@ -0,0 +1,10 @@
+-1.3208890577723018	0.025460321322479378	-0.1257927078878723	-1.3455206647658955	0.1504243148814659
```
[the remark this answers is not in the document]
gmrandazzo: True. removed. thanks!

catboost/pytest/test.py
```diff
@@ -90,6 +90,56 @@ def diff_tool(threshold=None):
...
+    output_calc_path = yatest.common.test_output_path('test.calc')
+    output_metric_path = yatest.common.test_output_path('test.metric')
+
+    cmd_fit = (
+        '--loss-function', 'MultiRMSEWithMissingValues',
+        '--boosting-type', boosting_type,
+        '-f', data_file('multiregression_with_missing', 'train'),
+        '-t', data_file('multiregression_with_missing', 'test'),
+        '--column-description', data_file('multiregression_with_missing', 'train.cd'),
+        '-i', '{}'.format(n_trees),
```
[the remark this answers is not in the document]
gmrandazzo: Fixed

catboost/libs/metrics/metric.cpp
```diff
@@ -544,6 +545,84 @@ TVector<TParamSet> TMultiRMSEMetric::ValidParamSets() {
...
+        return error;
+    };
+
+    return DispatchGenericLambda(evalImpl, !weight.empty(), !approxDelta.empty());
+}
+
+double TMultiRMSEWithMissingValues::GetFinalError(const TMetricHolder& error) const {
+    double finalError = 0.0;
+    for (size_t dim = 0; dim < error.Stats.size(); dim+=2) {
+	if (!FuzzyEquals(1+error.Stats[dim+1], 1+0.0, 1e-3)) {
```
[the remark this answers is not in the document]
gmrandazzo: Fixed.

catboost/libs/metrics/metric.cpp
```diff
@@ -544,6 +545,84 @@ TVector<TParamSet> TMultiRMSEMetric::ValidParamSets() {
...
+        
+        return error;
+    };
+
+    return DispatchGenericLambda(evalImpl, !weight.empty(), !approxDelta.empty());
+}
+
+double TMultiRMSEWithMissingValues::GetFinalError(const TMetricHolder& error) const {
+    double finalError = 0.0;
+    for (size_t dim = 0; dim < error.Stats.size(); dim+=2) {
```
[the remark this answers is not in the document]
gmrandazzo: Fixed.

catboost/private/libs/options/catboost_options.cpp
gmrandazzo: Strange. Is already there.

catboost/private/libs/target/data_providers.cpp
gmrandazzo: Fixed.

catboost/libs/metrics/optimal_const_for_loss.h
gmrandazzo: True. I revert it back to the original implementation.

catboost/libs/metrics/metric.cpp
gmrandazzo: ok. New proposal in master at the end of CheckPreprocessedTarget.

catboost/private/libs/target/data_providers.cpp
gmrandazzo: probably. Do you want to maintain or remove it? I mean, the time lost on this check should not be a limit factor on the overal catboost algorith.

catboost/libs/metrics/metric.cpp
```diff
@@ -544,6 +545,84 @@ TVector<TParamSet> TMultiRMSEMetric::ValidParamSets() {
...
+    TConstArrayRef<TConstArrayRef<float>> target,
+    TConstArrayRef<float> weight,
+    int begin,
+    int end
+) const {
+    const auto evalImpl = [=](bool useWeights, bool hasDelta) {
+        const auto realApprox = [=](int dim, int idx) { return approx[dim][idx] + (hasDelta ? approxDelta[dim][idx] : 0); };
+        const auto realWeight = [=](int idx) { return useWeights ? weight[idx] : 1; };
+        
+        TMetricHolder error(target.size()*2);
```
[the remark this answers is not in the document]
gmrandazzo: Fixed.

catboost/libs/metrics/metric.cpp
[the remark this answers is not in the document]
gmrandazzo: if(error.Stats[dim + 1] != 0) seems to work.

catboost/private/libs/algo_helpers/error_functions.h
```diff
@@ -218,6 +218,45 @@ class TMultiRMSEError final : public TMultiDerCalcer {
...
+    void CalcDers(
+        TConstArrayRef<double> approx,
+        TConstArrayRef<float> target,
+        float weight,
+        TVector<double>* der,
+        THessianInfo* der2
+    ) const override {
+        const int dim = target.size();
+        for (auto i : xrange(dim)) {
+	    //const float realWeight = weight * std::isnan(target[i]);
```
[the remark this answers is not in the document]
gmrandazzo: Fixed.

catboost/private/libs/algo_helpers/error_functions.h
```diff
@@ -218,6 +218,45 @@ class TMultiRMSEError final : public TMultiDerCalcer {
...
+        THessianInfo* der2
+    ) const override {
+        const int dim = target.size();
+        for (auto i : xrange(dim)) {
+	    //const float realWeight = weight * std::isnan(target[i]); 
+            //(*der)[i] = realWeight * (target[i] - approx[i]);
+	    if(std::isnan(target[i])) {
+	        (*der)[i] = 0.0;
+	    } else {
+		(*der)[i] = weight * (target[i] - approx[i]);
```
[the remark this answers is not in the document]
gmrandazzo: Fixed.

catboost/libs/metrics/metric.cpp
gmrandazzo: I wrote some implementation there... the current CheckProcessedTarget works very well and break when Nan are found. Now, the new version propose the possibility to skip Nans if MultiRMSEWithMissingValues or UserDefined metrics. It seems to be fine.

LyzhinIvan: Actually we have 2 functions with name `CheckPreprocessedTarget` - the first one is in [data_providers.cpp](https://github.com/catboost/catboost/pull/1680/files#diff-193809f26322d2a182596e906a6cfc5ed809066fb14c5fc68ba57cb96eb22449R100) and the second one is in this file - `metric.cpp`. I mean to add checks to the second one, because it is called for each metric, not for loss function only. Now if the user fit CatBoost with NaNs and `loss_function="MultiRMSEWithMissingValues"` and set `eval_metric="MultiRMSE"` then he get NaNs. It will be better to get error message instead of NaNs.

LyzhinIvan (changes requested):

catboost/private/libs/algo_helpers/error_functions.h
```diff
@@ -47,7 +47,7 @@ class IDerCalcer {
         return IsExpApprox;
     }
 
-    ui32 GetMaxSupportedDerivativeOrder() const {
+    i32 GetMaxSupportedDerivativeOrder() const {
```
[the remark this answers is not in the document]
gmrandazzo: No idea to be honest.
Come from the "merge" of the old mainline code.
Anyway now is back to ui32 as in the mainline source. 
Fixed.

catboost/libs/metrics/optimal_const_for_loss.h
```diff
@@ -29,6 +29,32 @@ namespace NCB {
         return targetSum / summaryWeight;
     }
 
+    inline float CalculateWeightedTargetAverageWithMissingValues(TConstArrayRef<float> target, TConstArrayRef<float> weights) {
+        TVector<float> target_without_nans;
```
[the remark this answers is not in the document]
gmrandazzo: right. Fixed.

catboost/libs/metrics/metric.cpp
```diff
@@ -5761,9 +5840,12 @@ TVector<THolder<IMetric>> CreateMetric(ELossFunction metric, const TLossParams&
         case ELossFunction::MultiRMSE:
             AppendTemporaryMetricsVector(TMultiRMSEMetric::Create(config), &result);
             break;
-        case ELossFunction::SurvivalAft:
-           AppendTemporaryMetricsVector(TSurvivalAftMetric::Create(config), &result);
+        case ELossFunction::MultiRMSEWithMissingValues:
+            AppendTemporaryMetricsVector(TMultiRMSEWithMissingValues::Create(config), &result);
             break;
+	case ELossFunction::SurvivalAft:
```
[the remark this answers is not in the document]
gmrandazzo: fixed.

catboost/private/libs/target/data_providers.cpp
```diff
@@ -25,8 +25,16 @@
 
 namespace NCB {
 
-    inline bool IsSafeTarget(float value) {
-        return Abs(value) < 1e6f;
+    inline bool IsSafeTarget(double value, bool allowNan) {
+        if (allowNan) {
```
LyzhinIvan, line 29: I suggest to make it more elegant:
```
return value < 1e6f || (allowNan && IsNan(value));
```

catboost/private/libs/options/catboost_options.cpp
LyzhinIvan: At 677-678 lines message "You can use boost_from_average ... , MultiRMSE **or MultiRMSEWithMissingValues**."

catboost/libs/metrics/metric.cpp
```diff
@@ -6387,6 +6471,12 @@ void CheckPreprocessedTarget(
                   "metric/loss-function " << lossFunction << " is a Multiclassification metric, "
                   " each target label should be a nonnegative integer");
     }
+
+    if (lossFunction == ELossFunction::MultiRMSE) {
```
LyzhinIvan, line 6475: It will be better to write `if (lossFunction != ELossFunction::MultiRMSEWithMissingValues) {` here.

catboost/libs/metrics/optimal_const_for_loss.h
```diff
@@ -29,6 +29,32 @@ namespace NCB {
         return targetSum / summaryWeight;
     }
 
+    inline float CalculateWeightedTargetAverageWithMissingValues(TConstArrayRef<float> target, TConstArrayRef<float> weights) {
+        TVector<float> targetWithoutNans;
```
LyzhinIvan, line 33: `CalculateWeightedTargetAverage` is very simple function and it will be ok if you just copy it here with small modifications for NaNs. It will be more short and without target and weights copying.

catboost/libs/metrics/optimal_const_for_loss.h
```diff
@@ -105,6 +131,10 @@ namespace NCB {
         switch(lossFunction) {
             case ELossFunction::RMSE:
                 return CalculateWeightedTargetAverage(target, weights);
+            //case ELossFunction::MultiRMSEWithMissingValues:
```
LyzhinIvan, line 134: remove these commented lines

catboost/libs/data/target.cpp
```diff
@@ -417,6 +417,24 @@ ERawTargetType TRawTargetDataProvider::GetTargetType() const {
     return Data.TargetType;
 }
 
+
+bool TRawTargetDataProvider::IsTargetsContainsNan() const {
```
LyzhinIvan, line 421: it is not required anymore

LyzhinIvan (changes requested):

catboost/libs/metrics/optimal_const_for_loss.h
gmrandazzo: Done.

catboost/private/libs/target/data_providers.cpp
```diff
@@ -25,8 +25,16 @@
 
 namespace NCB {
 
-    inline bool IsSafeTarget(float value) {
-        return Abs(value) < 1e6f;
+    inline bool IsSafeTarget(double value, bool allowNan) {
```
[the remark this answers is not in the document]
gmrandazzo: Done

catboost/private/libs/target/data_providers.cpp
```diff
@@ -25,8 +25,16 @@
...
+        if (allowNan) {
+            if (IsNan(value)) {
+                return true;
+            } else{
+                return Abs(value) < 2e6f;
```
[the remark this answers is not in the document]
gmrandazzo: Done

catboost/private/libs/target/data_providers.cpp
gmrandazzo: You are right. Better.
Done.

catboost/libs/metrics/metric.cpp
gmrandazzo: ok.
Done

catboost/libs/metrics/optimal_const_for_loss.h
gmrandazzo: Ok. It makes sense also to me. 
Done

catboost/libs/data/target.cpp
gmrandazzo: True! Now is much more elegant!
Thank you @LyzhinIvan 
Done.

catboost/private/libs/target/data_providers.cpp
gmrandazzo: ok.
Done

catboost/libs/metrics/metric.cpp
```diff
@@ -544,6 +545,84 @@ TVector<TParamSet> TMultiRMSEMetric::ValidParamSets() {
     return {TParamSet{{TParamInfo{"use_weights", false, true}}, ""}};
 };
 
+/* MultiRMSEWithMissingValues */
+namespace {
+    struct TMultiRMSEWithMissingValues final: public TAdditiveMultiRegressionMetric {
+        explicit TMultiRMSEWithMissingValues(const TLossParams& params)
+           : TAdditiveMultiRegressionMetric(ELossFunction::MultiRMSE, params)
```
LyzhinIvan, line 552: ELossFunction::MultiRMSEWithMissingValues

LyzhinIvan (changes requested):

catboost/libs/metrics/metric.cpp
gmrandazzo: right... thank you @LyzhinIvan 
Done.

LyzhinIvan (approved):

LyzhinIvan:
@arcadia-devtools Ship it!

arcadia-devtools:
@gmrandazzo, please agree to the CLA (https://github.com/catboost/catboost/blob/master/CONTRIBUTING.md#provide-contributions)

arcadia-devtools:
@LyzhinIvan, internal review request created: [1849441](https://a.yandex-team.ru/review/1849441)

catboost/private/libs/target/data_providers.cpp
```diff
@@ -494,8 +494,11 @@ namespace NCB {
                 << " loss/metrics require target data"
             );
         }
+        
+        bool needCheckTarget = !mainLossFunction || !IsRegressionObjective(mainLossFunction->GetLossFunction());
+        //Acept nan in MultiRMSEWithMissingValues
+        bool allowNanTarget = (mainLossFunction->GetLossFunction() == ELossFunction::MultiRMSEWithMissingValues);
```
LyzhinIvan, line 500: Sometime model does not contain loss function (for example, after `sum_models`), now it fails. Let's allow NaNs in such case.
`bool allowNanTarget = !mainLossFunction || (mainLossFunction->GetLossFunction() == ELossFunction::MultiRMSEWithMissingValues);
`

catboost/private/libs/target/data_providers.cpp
```diff
@@ -494,8 +494,11 @@ namespace NCB {
                 << " loss/metrics require target data"
             );
         }
+        
+        bool needCheckTarget = !mainLossFunction || !IsRegressionObjective(mainLossFunction->GetLossFunction());
+        //Acept nan in MultiRMSEWithMissingValues
```
LyzhinIvan, line 499: `// Accept...`

catboost/private/libs/target/data_providers.cpp
```diff
@@ -25,11 +25,10 @@
 
 namespace NCB {
 
-    inline bool IsSafeTarget(float value) {
-        return Abs(value) < 1e6f;
+    inline bool IsSafeTarget(float value, bool allowNan) {
+        return value < 1e6f || (allowNan && IsNan(value));
```
LyzhinIvan, line 29: we forgot `Abs(value)` here

catboost/libs/metrics/metric.cpp
```diff
@@ -544,6 +545,84 @@ TVector<TParamSet> TMultiRMSEMetric::ValidParamSets() {
...
+    const auto evalImpl = [=](bool useWeights, bool hasDelta) {
+        const auto realApprox = [=](int dim, int idx) { return approx[dim][idx] + (hasDelta ? approxDelta[dim][idx] : 0); };
+        const auto realWeight = [=](int idx) { return useWeights ? weight[idx] : 1; };
+        
+        TMetricHolder error(target.size() * 2);
+        for (auto dim : xrange(target.size())) {
+            double sumWeights = 0.0;
+            double sumErrors = 0.0;
+            for (auto i : xrange(begin, end)) {
+                if (!std::isnan(target[dim][i])) {
```
LyzhinIvan, line 593: `std::isnan` -> `IsNan`

catboost/libs/metrics/metric.cpp
```diff
@@ -544,6 +545,84 @@ TVector<TParamSet> TMultiRMSEMetric::ValidParamSets() {
...
+        return error;
+    };
+
+    return DispatchGenericLambda(evalImpl, !weight.empty(), !approxDelta.empty());
+}
+
+double TMultiRMSEWithMissingValues::GetFinalError(const TMetricHolder& error) const {
+    double finalError = 0.0;
+    for (size_t dim = 0; dim < error.Stats.size(); dim += 2) {
+        if(error.Stats[dim + 1] != 0) {
```
LyzhinIvan, line 611: space after if: `if (`

catboost/libs/metrics/metric.cpp
```diff
@@ -544,6 +545,84 @@ TVector<TParamSet> TMultiRMSEMetric::ValidParamSets() {
...
+    };
+
+    return DispatchGenericLambda(evalImpl, !weight.empty(), !approxDelta.empty());
+}
+
+double TMultiRMSEWithMissingValues::GetFinalError(const TMetricHolder& error) const {
+    double finalError = 0.0;
+    for (size_t dim = 0; dim < error.Stats.size(); dim += 2) {
+        if(error.Stats[dim + 1] != 0) { 
+            finalError += error.Stats[dim]/error.Stats[dim+1];
```
LyzhinIvan, line 612: spaces around operators `error.Stats[dim] / error.Stats[dim+1];`

catboost/libs/metrics/metric.cpp
```diff
@@ -6387,6 +6471,12 @@ void CheckPreprocessedTarget(
                   "metric/loss-function " << lossFunction << " is a Multiclassification metric, "
                   " each target label should be a nonnegative integer");
     }
+
+    if (lossFunction != ELossFunction::MultiRMSEWithMissingValues) {
+        for (auto objectIdx : xrange(target.size())){
+            CB_ENSURE(!std::isnan(target[objectIdx]), "metric/loss-function " << lossFunction << " do not allows nan value on target");
```
LyzhinIvan, line 6477: `std::isnan` -> `IsNan`

catboost/private/libs/algo_helpers/error_functions.h
```diff
@@ -218,6 +218,43 @@ class TMultiRMSEError final : public TMultiDerCalcer {
...
+    void CalcDers(
+        TConstArrayRef<double> approx,
+        TConstArrayRef<float> target,
+        float weight,
+        TVector<double>* der,
+        THessianInfo* der2
+    ) const override {
+        const int dim = target.size();
+        for (auto i : xrange(dim)) {
+            if(std::isnan(target[i])) {
```
LyzhinIvan, line 236: `if (`

LyzhinIvan: `IsNan`

catboost/private/libs/algo_helpers/error_functions.h
```diff
@@ -218,6 +218,43 @@ class TMultiRMSEError final : public TMultiDerCalcer {
...
+                (*der)[i] = weight * (target[i] - approx[i]);
+            }
+        }
+
+        if (der2 != nullptr) {
+            Y_ASSERT(der2->HessianType == EHessianType::Diagonal &&
+                     der2->ApproxDimension == dim);
+
+            for (auto i : xrange(dim)) {
+                if (std::isnan(target[i])) {
```
LyzhinIvan, line 248: `IsNan`

catboost/private/libs/target/data_providers.cpp
```diff
@@ -494,8 +494,11 @@ namespace NCB {
                 << " loss/metrics require target data"
             );
         }
+        
+        bool needCheckTarget = !mainLossFunction || !IsRegressionObjective(mainLossFunction->GetLossFunction());
```
LyzhinIvan, line 498: `const bool`

catboost/private/libs/target/data_providers.cpp
LyzhinIvan: `const bool`

LyzhinIvan (changes requested):

catboost/pytest/test.py
```diff
@@ -90,6 +90,56 @@ def diff_tool(threshold=None):
...
+        '-o', output_metric_path,
+        '--metrics', 'MultiRMSEWithMissingValues'
+    )
+    yatest.common.execute(cmd_metric)
+    return [
+        local_canonical_file(output_eval_path),
+        local_canonical_file(output_calc_path),
+        local_canonical_file(output_metric_path)
+    ]
+
```
LyzhinIvan, line 142: pep8 requires 2 blank lines between functions

catboost/private/libs/target/data_providers.cpp
gmrandazzo: I see. 
Done.

catboost/private/libs/target/data_providers.cpp
gmrandazzo: typo fixed.

catboost/private/libs/target/data_providers.cpp
gmrandazzo: uh..,.right.. done

catboost/libs/metrics/metric.cpp
gmrandazzo: Done

catboost/libs/metrics/metric.cpp
gmrandazzo: Done

catboost/libs/metrics/metric.cpp
gmrandazzo: Done

catboost/private/libs/algo_helpers/error_functions.h
gmrandazzo: Done

catboost/private/libs/algo_helpers/error_functions.h
gmrandazzo: Done

catboost/private/libs/target/data_providers.cpp
gmrandazzo: Done

catboost/pytest/test.py
gmrandazzo: ok. Done

LyzhinIvan:
@arcadia-devtools Ship it!

arcadia-devtools:
CLA already signed

LyzhinIvan (approved):

arcadia-devtools:
@LyzhinIvan, internal review request updated: [1849441](https://a.yandex-team.ru/review/1849441)

Issue #1592, opened 2021-02-20
Node.js interface of CatBoost


nikitxskv:
Ability to make predictions of CatBoost model  in Node.js.
Similar functionality is available for XGBoost: https://github.com/nuanio/xgboost-node
CatBoost C Api: https://github.com/catboost/catboost/blob/master/catboost/libs/model_interface/c_api.h#L82

nikitxskv:
Done: https://www.npmjs.com/package/catboost

PR #1681, opened 2021-04-28
[node-package] Handle a call to predict() on not loaded model 

IKholopov:
1. Fixing a bug where calling predict() on the empty model that hasn't been loaded results in the infinite loop execution.
2. More careful error-handling: NHelper::Check... checks were only scheduling the exception without interrupting the flow of execution, which in turn could cause UB. Handling the result of the checks and returning early. Why C-like (or go-like?) error handling instead of using some custom exception? Node package has to be built on each host that would like to install it, so every client should have a compiler version capable of compiling the bindings. Catboost exceptions make use of C++17 features and not work well with some common compilers (like GCC), so it is much safer to add a couple of if statements.

### DIFF
--- a/catboost/node-package/bindings/tests/model.ts
+++ b/catboost/node-package/bindings/tests/model.ts
@@ -106,4 +106,10 @@ function testLoadNonExistentOnConstruction() {
 }
 assert.throws(testLoadNonExistentOnConstruction);
 
+function testPredictOnEmptyModel() {
+    const model = new Model();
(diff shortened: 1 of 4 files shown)


IKholopov:
#1592

@nikitxskv

arcadia-devtools:
CLA already signed

arcadia-devtools:
@nikitxskv, internal review request created: [1770953](https://a.yandex-team.ru/review/1770953)

PR #1695, opened 2021-05-09
Explicitly specify WHEEL 'Root-Is-Purelib' value

thundergolfer:
Version `0.25.1` (latest) has the following WHEEL file: 

```bash
cat catboost-0.25.1.dist-info/WHEEL                                                                                                                                                                                                                                                                                 
Wheel-Version: 1.0
```

This causes problems for package installation code that assumes the `Root-Is-Purelib` key is present in the WHEEL. An example is this bug: https://github.com/bazelbuild/rules_python/issues/435.

The `Root-Is-Purelib` key is described in ["PEP-0427-- The Wheel Binary Package Format 1.0"](https://www.python.org/dev/peps/pep-0427/#id14).

Now there's certainly an argument to be made that the `Root-Is-Purelib` is not mandatory and its absence shouldn't be an error, but in any case it's good to be explicit.

### DIFF
--- a/catboost/python-package/catboost.dist-info/WHEEL
+++ b/catboost/python-package/catboost.dist-info/WHEEL
@@ -1,1 +1,2 @@
 Wheel-Version: 1.0
+Root-Is-Purelib: false


kizill:
Thanks for your help! We will publish wheel with this fix in next release by the end of next week.

kizill:
@arcadia-devtools Ship it!

arcadia-devtools:
@thundergolfer, please agree to the CLA (https://github.com/catboost/catboost/blob/master/CONTRIBUTING.md#provide-contributions)

arcadia-devtools:
@kizill, internal review request created: [1776864](https://a.yandex-team.ru/review/1776864)

PR #1807, opened 2021-08-13
Focal loss c++ implementation

diditforlulz273:
I hereby agree to the terms of the CLA available at [https://yandex.ru/legal/cla/?lang=en].

Adds a focal loss CPU implementation. Focal loss is widely used in classification tasks with highly imbalanced data.
Details can be found here:
[https://arxiv.org/abs/1708.02002](https://arxiv.org/abs/1708.02002) - original paper
[https://paperswithcode.com/method/focal-loss](https://paperswithcode.com/method/focal-loss) - 216 papers used it.

Possible problems:
1. Never tested in GPU mode. (Although on CPU is being used already).
2. Alpha and Gamma parameters are passed by user to loss, but hardcoded in metric. If the way I did it in loss is correct, I'll add it to the metric too.
3. Not covered by tests.
4. No documentation is written, intended use is through Python Catboost wrapper with parameters like
 "loss_function": "Focal:focal_alpha=0.75;focal_gamma=2.0"


catboost/libs/metrics/metric.cpp
```diff
@@ -1779,6 +1779,128 @@ TVector<TParamSet> TTweedieMetric::ValidParamSets() {
...
+    Y_ASSERT(!isExpApprox);
+    const auto impl = [=] (auto hasDelta, auto hasWeight, TConstArrayRef<double> approx, TConstArrayRef<double> approxDelta) {
+        TMetricHolder error(2);
+        for (int k : xrange(begin, end)) {
+            double curApprox = approx[k];
+            if (hasDelta) {
+                curApprox += approxDelta[k];
+            }
+            const float w = hasWeight ? weight[k] : 1;
+            double FocalAlpha = 0.75;
```
sagol, line 1847: It is necessary to pass these parameters into the metric somehow.

diditforlulz273: I'd better wait for maintainers' advice on how to do it in the most correct way(probably in the way I did it with loss params)

catboost/libs/metrics/metric.cpp
```diff
@@ -1779,6 +1779,128 @@ TVector<TParamSet> TTweedieMetric::ValidParamSets() {
...
+            double FocalGamma = 2;
```
sagol, line 1848: This parameter too.

kizill:
@arcadia-devtools Ship it!

langkkl:
thanks your great job.
for catboost 1.0.6,metric.cpp should add a GetDescription method.

andrey-khropov:
Thank you for your contribution. Sorry for the long wait. I've merged this PR (the original author is preserved). 
I'll add tests.

Issue #1851, opened 2021-09-17
R version might compile with incorrect headers


david-cortes:
The R version for CatBoost is compiled through an atypical build system which bundles R headers taken from version 3.5.0. Currently, R is at version 4.1, and these headers are not meant to be unchanging between versions (R packages are supposed to compile for a specific R version, and when using the R build system, they take system headers from the current R installation).

In other packages, this practice has lead to very weird and seemingly random errors (including segmentation faults and very incorrect results) - see for example https://github.com/microsoft/LightGBM/issues/4216

Would be better to update these headers to at least 4.0 and require a minimum R version for the package (configurable through the `DESCRIPTION` file at the root of the R package directory).

PR #1852, opened 2021-09-17
[R] Fix memory safety issues

david-cortes:
This PR fixes a couple of potential memory leaks and use-after-free errors in the R version. In short, lots of these errors come from 2 patterns:
* Throwing R errors in a catch block, which trigger C long jumps and end up leaking the exception object.
* Not protecting R objects of dynamic storage, which might get deallocated at the moment they are to be used.

Some information about the new functions used here (particularly unwind protection) can be found in the R extensions manual: https://cran.r-project.org/doc/manuals/R-exts.html

**IMPORTANT:** Getting this PR to compile requires updating the R headers to something more recent, such as version 4.0.0 (see https://github.com/catboost/catboost/issues/1851). Current headers under `contrib` are at 3.5.0 and lack some of the functions used in this PR.

### DIFF
--- a/catboost/R-package/src/catboostr.cpp
+++ b/catboost/R-package/src/catboostr.cpp
@@ -30,32 +30,68 @@
 #include <util/system/info.h>
 
 #include <algorithm>
+#include <cstdio>
 
 #if defined(SIZEOF_SIZE_T)
 #undef SIZEOF_SIZE_T
 #endif
 
+#include <R.h>
 #include <Rinternals.h>
-
 
 using namespace NCB;
 
 
+#define MAXLEN_R_ERRMSG 1024
+char buffer_R_errmsg[MAXLEN_R_ERRMSG];
+void copy_exception_msg_to_buffer(std::exception &e) {
+    std::snprintf(buffer_R_errmsg, MAXLEN_R_ERRMSG, "%s\n", e.what());
+}
+
+struct CB_R_ERROR {SEXP uw_token;};
+
+void loggingFunc(const char* str, size_t len, TCustomLoggingObject) {
+    TString slicedStr(str, 0, len);
+    Rprintf("%s", slicedStr.c_str());
+    R_FlushConsole();
+}
+
+void loggingErrFunc(const char* str, size_t len, TCustomLoggingObject) {
+    TString slicedStr(str, 0, len);
+    REprintf("%s", slicedStr.c_str());
+    R_FlushConsole();
+}
+
+class CB_R_context {
+public:
+    CB_R_context() {
+        SetCustomLoggingFunction(loggingFunc, loggingErrFunc);
+    }
+
+    ~CB_R_context() {
+        RestoreOriginalLogger();
+    }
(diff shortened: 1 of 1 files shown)


kizill:
Is it safe to upgrade headers and minimal requirements to 4.0.0 ? Do you know if there are any stats about version distribution in the wild?

david-cortes:
If you mean to ask "is it safe to compile with R headers 4.0.0 and load in a different R version", then no, there's no guarantee that it will work correctly, same as it isn't safe to compile with R headers 3.5.0 and then load in R 4.0.0 like it's currently being done (https://github.com/catboost/catboost/issues/1851). AFAIK it's only safe to reuse headers between "patch" versions, like 4.0.0 -> 4.0.1. The R headers are only meant to be stable in the sense that, if including them from an R installation and using the functions they contain, the functions and signatures from the headers will keep working in newer R installations, but the underlying definitions that they use internally might change - for example, they use type `SEXP` and the functions that return `SEXP` have remained the same for many versions, but the structure behind a `SEXP` type AFAIK has changed. Some of the headers are also setup-specific - for example, the visibility header will look different in windows, linux, and mac, and if building R with a non-default compiler it might also look different. The only other libraries that I know of which provide independent binary R builds typically build them for a specific R version and platform (e.g. https://anaconda.org/conda-forge/r-arrow/files).

If you mean to ask "is it common for users to have R version 4.0.0 or greater", then I'd guess it more or less is, since it was released 17 months ago and the current version is 4.1.1. R 4.0 is currently available in debian stable, but not in ubuntu LTS (unless using R's own repository instead of ubuntu's). In CRAN servers, they are not currently testing packages against anything older than R 4.0.4.

I don't know of any reliable source for counting users of R versions, but a check for number of R downloads in the past week from `cranlogs` shows that roughly 94% of downloads were for R>=4.0.0:
```r
library(cranlogs)
library(data.table)
df = cran_downloads("R")
setDT(df)
df[
    grepl("^\\d+\\.\\d+\\.\\d+.*$", version)
][,
  .(count, tot = sum(count), version = gsub("^(\\d+\\.\\d+)\\.\\d+.*$", "\\1", version))
][,
   .(pct = sum(count) / max(tot))
  , by=.(version)
][
    order(-pct)
    , .(version, pct = sprintf("%.2f%%", 100*pct))
][1:5]
```
```
   version    pct
1:     4.1 87.93%
2:     4.0  6.48%
3:     3.6  3.55%
4:     3.5  0.52%
5:     3.3  0.36%
```

Although I think this PR might compile with headers from R 3.6.0 too, just not with 3.5.0.

kizill:
@arcadia-devtools Ship it!

kizill:
I'll try to do something with R headers in our repo, thanks. I guess i'll try 3.6 first.

arcadia-devtools:
CLA already signed

arcadia-devtools:
@kizill, internal review request created: [2049780](https://a.yandex-team.ru/review/2049780)

PR #1905, opened 2021-11-04
First implementation without test, which add support for loading larg…

CamilYed:
It's only draft, I need to add all required tests

I will be grateful for your help :)

### DIFF
--- a/catboost/jvm-packages/catboost4j-prediction/src/main/java/ai/catboost/CatBoostModel.java
+++ b/catboost/jvm-packages/catboost4j-prediction/src/main/java/ai/catboost/CatBoostModel.java
@@ -6,6 +6,7 @@ import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
 
+import java.nio.ByteBuffer;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Comparator;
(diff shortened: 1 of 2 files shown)


catboost/jvm-packages/catboost4j-prediction/src/main/java/ai/catboost/CatBoostModel.java
```diff
@@ -233,6 +234,36 @@ public static CatBoostModel loadModel(final InputStream in) throws CatBoostError
...
+
+        while (in.read(copyBufferArray) != -1) {
+            memoryForModel.put(copyBufferArray);
+        }
+
+        implLibrary.catBoostLoadModelFromArray(memoryForModel.array(), handles);
+        return new CatBoostModel(handles[0]);
+    }
+
+    private static int countAllBytes(final InputStream in) throws IOException {
```
CamilYed, line 262: It's not the best idea, because after this operation we have to start read stream from beginning. 
Not every stream support reset operation.

CamilYed:
How to run tests on Osx?
image

andrey-khropov:
> How to run tests on Osx? image

Running `mvn test` from `catboost4j-prediction` source code directory works for me on macOS. Might be some issue with the way your IDE interprets Maven's `pom` files.

PR #2161, opened 2022-09-08
Refactor crate avoid build

jlloh:
Originally had this MR to fix the build for Mac M1. https://github.com/catboost/catboost/pull/2160 .

After testing this in my environment, I realised that with the current implementation:
1. The current implementation does not allow the crate to be easily published to crates.io due to dependencies on the base repo as it compiles the shared library from source
  * This repo is actually pretty big, and if caching this for CI, it is >1GB and affects build speed pretty significantly
3. Looking at how onnxruntime's [rust bindings ](https://github.com/nbigaouette/onnxruntime-rs/blob/master/onnxruntime-sys/build.rs#L23)and pytorch's rust bindings download the shared library (in this case libcatboost.so/dylib) over building it
4. Taking inspiration from that, I modified catboost-sys to download the libcatboost shared library from github releases (and did a soft link to `../../libs/model_interface` to allow bindgen to generate the necessary rust bindings. I managed to publish the crate [here](https://crates.io/crates/catboost-sys/0.1.2). 
5. After that, I subsequently changed the dependencies for the actual crate itself and published it to [here](https://crates.io/crates/catboost-rs/0.1.1).
   * Note that I had to call the crate catboost-rs instead of catboost because there's an existing empty crate already using that name, preventing me from using it

Would like to get feedback on whether this is a reasonable approach, and can be merged into master.

### DIFF
--- a/catboost/rust-package/Cargo.lock
+++ b/catboost/rust-package/Cargo.lock
@@ -69,7 +69,7 @@ dependencies = [
 
 [[package]]
 name = "catboost-sys"
-version = "0.1.0"
+version = "0.1.1"
 dependencies = [
  "bindgen",
 ]
--- a/catboost/rust-package/Cargo.lock
+++ b/catboost/rust-package/Cargo.lock
@@ -3,6 +3,12 @@
 version = 3
 
 [[package]]
+name = "adler"
+version = "1.0.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
+
+[[package]]
 name = "aho-corasick"
 version = "0.7.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


jlloh:
Minor update: As I haven't heard back from the maintainers, and it seems like overkill to fork the whole repo and maintain it, I have opted to split out the code to it's own [repository](https://github.com/jlloh/catboost-rs), similar to onnxruntime-rs, since all I need are the files in `catboost/libs/model_interface` for `catboost-sys`. 

Instead, to build `catboost-sys`, the user is responsible for cloning the catboost repo himself and set an environment variable `CATBOOST_DIR`.

The crate is published [here](https://crates.io/crates/catboost-rs).

Leaving the MR open for now in case the maintainers ever see it.

marcomq:
Just as info, 

I also created a branch to avoid catboost builds, but made it an optional feature. By default, it will keep the current build behavior:
https://github.com/catboost/catboost/compare/master...marcomq:catboost:master

Not sure yet if I keep the xgb_compat file. I added it to simulate xgboost function signature, before I figured out that the predict is completely different in catboost.

Unfortunately, I have other stuff to do now :/ 
It will probably take me at least 2 weeks until I have time to add the changes to a sample project, train a proper model and test everything locally.
Will probably create a PR then. Currently, I only verified the changes by unit tests.

PR #2498, opened 2023-09-21
Docs: Add a missing colon to the custom_metrics example.

lmacken:
This simply adds a `:` to the example: `Quantile:alpha=0.1`


andrey-khropov:
Thanks for fix, it is actually a colon, not a semilcolon.

PR #2500, opened 2023-09-22
Add a missing colon to the custom_metrics example in the docs

lmacken:
This simply adds a `:` to the example: `Quantile:alpha=0.1`


lmacken:
My previous PR was merged after all, so this one is not needed.

PR #2537, opened 2023-11-23
CatBoostModel.java: add `byte[]` interface

catap:
Here a trivial changes which can be treated as public domain.

Goal: to avoid wrap `byte[]` into stream which will be copied into `byte[]` to be used inside.

### DIFF
--- a/catboost/jvm-packages/catboost4j-prediction/src/main/java/ai/catboost/CatBoostModel.java
+++ b/catboost/jvm-packages/catboost4j-prediction/src/main/java/ai/catboost/CatBoostModel.java
@@ -225,6 +225,34 @@ public class CatBoostModel implements AutoCloseable {
     }
 
     /**
+     * Load CatBoost model from byte array.
+     *
+     * @param model Byte array containing model.
+     * @return      CatBoost model.
+     * @throws CatBoostError When failed to load model.
+     */
+    @NotNull
+    public static CatBoostModel loadModel(final @NotNull byte[] model) throws CatBoostError {
+        return loadModel(modelPath, "bin");
+    }
+
+    /**
+     * Load CatBoost model from file modelPath.
+     *
+     * @param model       Byte array containing model.
+     * @param modelFormat Model file format (bin or json)
+     * @return            CatBoost model.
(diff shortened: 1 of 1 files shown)


catboost/jvm-packages/catboost4j-prediction/src/main/java/ai/catboost/CatBoostModel.java
```diff
@@ -224,6 +224,34 @@ public int compare(Feature v1, Feature v2) {
...
+    /**
+     * Load CatBoost model from byte array.
+     *
+     * @param model Byte array containing model.
+     * @return      CatBoost model.
+     * @throws CatBoostError When failed to load model.
+     */
+    @NotNull
+    public static CatBoostModel loadModel(final @NotNull byte[] model) throws CatBoostError {
+        return loadModel(modelPath, "bin");
```
andrey-khropov, line 236: `modelPath` -> `serializedModel`

andrey-khropov (changes requested):
Please add a test to CatBoostModelTest.java. You can check that model returns the same predictions when loaded directly from a file or when read from a file to a byte array and then loaded through your new function.

You can find sample model files and features data to use with these models in other tests in that file.

andrey-khropov:
> Here a trivial changes which can be treated as public domain.

I can ask our lawyers whether we can accept such terms but it would be easier if you just agree to [the CLA](https://yandex.ru/legal/cla/)

catap:
> > Here a trivial changes which can be treated as public domain.
> 
> I can ask our lawyers whether we can accept such terms but it would be easier if you just agree to [the CLA](https://yandex.ru/legal/cla/)

Or I may close that PR and ask you to re-implement it to make things faster? :)

andrey-khropov:
> > > Here a trivial changes which can be treated as public domain.
> > 
> > 
> > I can ask our lawyers whether we can accept such terms but it would be easier if you just agree to [the CLA](https://yandex.ru/legal/cla/)
> 
> Or I may close that PR and ask you to re-implement it to make things faster? :)

Ok. https://github.com/catboost/catboost/issues/2539

Issue #2243, opened 2022-12-07
Model predictions of type str when training labels are boolean


mkleinbort-ic:
Problem:

When I fit a `CatBoostClassifier` on a boolean target (using a pd.Series with True/False entries), the resulting model returns the strings `'True'`/`'False'` as the predicted labels - not `True`/`False`

Is there a reason for this? It introduces some annoying glue-code into my applications.

catboost version: 1.1.1

Operating System: Windows/Mac

Evgueni-Petrov-aka-espetrov:
duplicate of https://github.com/catboost/catboost/issues/1954

Issue #1954, opened 2021-12-14
[Python] Predicted target type is string but is expected to be boolean


memo26167:
Problem: The training target is boolean and the predicted target is string. When calling the score method, it throws an error.
catboost version: 1.0.3
Operating System: Windows 10
CPU: -
GPU: -

```python
from catboost import CatBoostClassifier
import numpy as np

clf = CatBoostClassifier(silent=True)
clf.fit(np.array([[0,1,2],[2,3,4]]), np.array([[True],[False]]))
print(clf.predict(np.array([[0,1,2],[2,3,4]])))

clf.score(np.array([[0,1,2],[2,3,4]]), np.array([[True],[False]]))
```

Throwing the following error
```
CatBoostError                             Traceback (most recent call last)
----> 1 clf.score(np.array([[0,1,2],[2,3,4]]), np.array([[True],[False]]))

\my_env\lib\site-packages\catboost\core.py in score(self, X, y)
   5032                 raise CatBoostError('predicted classes have string type but specified y is numeric')
   5033             elif np.issubdtype(y.dtype, np.bool_):
-> 5034                 raise CatBoostError('predicted classes have string type but specified y is boolean')
   5035         return np.mean(np.array(predicted_classes) == np.array(y))
   5036 

CatBoostError: predicted classes have string type but specified y is boolean
```

PR #2538, opened 2023-11-25
[1954] Use bool type for boolean predictions

NickVeld:
Note: For now it is a PoC that solves the issues. It depends on the further collaboration how good the problem solution can be developed in the end.

Solves #1954 #2243


andrey-khropov:
Unfortunately this solution is too crude and will break a valid case of class labels that contain `True`, `False` strings. I'm in the process making the proper fix which is unfortunately more involved. Sorry for a delayed response.

NickVeld:
It would be waste of time to create proper solution and after several months to get reply that the fix is in progress somewhere else ;)

andrey-khropov:
> It would be waste of time to create proper solution and after several months to get reply that the fix is in progress somewhere else ;)

I'm sorry, we should had replied sooner. But the proper fix have been more complicated and took time to investigate.

PR #2568, opened 2024-01-11
Add functions for getting indices of features of different types to C API

nimusp:

I am adding methods for retrieving feature indices.
The Python API already has it, but the C API is missing it.

### DIFF
--- a/catboost/libs/model/model.cpp
+++ b/catboost/libs/model/model.cpp
@@ -1341,6 +1341,46 @@ TVector<TString> GetModelUsedFeaturesNames(const TFullModel& model) {
     return result;
 }
 
+TVector<int> GetModelCatFeaturesIndices(const TFullModel& model) {
+    const auto features = model.ModelTrees->GetCatFeatures();
+    TVector<int> indices;
+    indices.reserve(features.size());
+    for (const auto& feature : features) {
+        indices.push_back(feature.Position.FlatIndex);
+    }
+    return indices;
+}
+
+TVector<int> GetModelFloatFeaturesIndices(const TFullModel& model) {
+    const auto features = model.ModelTrees->GetFloatFeatures();
+    TVector<int> indices;
+    indices.reserve(features.size());
+    for (const auto& feature : features) {
+        indices.push_back(feature.Position.FlatIndex);
+    }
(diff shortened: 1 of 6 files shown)


catboost/libs/model/model.cpp
```diff
@@ -1341,6 +1341,46 @@ TVector<TString> GetModelUsedFeaturesNames(const TFullModel& model) {
...
+
+TVector<int> GetModelEmbeddingFeaturesIndices(const TFullModel& model) {
+    const auto features = model.ModelTrees->GetEmbeddingFeatures();
+    TVector<int> indices;
+    indices.reserve(features.size());
+    for (const auto& feature : features) {
+        indices.push_back(feature.Position.FlatIndex);
+    }
+    return indices;
+}
```
andrey-khropov, line 1382: Please avoid copy-paste and create a helper function that will create TVector<int> from the result of Get*Features functions.

catboost/libs/model_interface/c_api.cpp
```diff
@@ -888,5 +888,61 @@ CATBOOST_API bool GetModelUsedFeaturesNames(ModelCalcerHandle* modelHandle, char
...
+    if (!*indices) {
+        return false;
+    }
+
+    *count = featureIndices.size();
+    for (size_t i = 0; i != featureIndices.size(); ++i) {
+        (*indices)[i] = featureIndices[i];
+    }
+    return true;
+}
```
andrey-khropov, line 945: Please avoid copy-paste and create a helper function that will copy the data from TVector to `indices` and `count`.

andrey-khropov, line 945: Please wrap GetModel*FeaturesIndices call in try { } catch block and set an error message and return false if the exception is cached (see similar logic in other functions in this file).

andrey-khropov (changes requested):

catboost/libs/model/model.cpp
nimusp: There may be a problem with converting to the parent (TBaseArrayFeature) when using TConstArrayRef, so I added a workaround with templates

andrey-khropov (approved):

Issue #2580, opened 2024-01-25
Native build requires MSVS on MacOS


jnovikov:
Problem: Native build requires MSVS on MacOS
catboost version: master
Operating System: MacOS
CPU: ARM64 (M1)
GPU: -

While building the rust package from master branch I faced next issue:

```
Caused by:
  process didn't exit successfully: `/target/debug/build/catboost-sys-4b2feec9828e6014/build-script-build` (exit status: 101)
  --- stderr
  INFO:root:target_platform=darwin-arm64. Building targets catboostmodel with PIC
  Traceback (most recent call last):
    File "/Users/jnovikov/.cargo/git/checkouts/catboost-113273d315325944/3410894/catboost/rust-package/catboost-sys/../../../build/build_native.py", line 594, in <module>
      build(Opts(**vars(parsed_args)))
    File "/Users/jnovikov/.cargo/git/checkouts/catboost-113273d315325944/3410894/catboost/rust-package/catboost-sys/../../../build/build_native.py", line 493, in build
      msvs_dir = get_msvs_dir(opts.msvs_installation_path, opts.msvs_version)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/Users/jnovikov/.cargo/git/checkouts/catboost-113273d315325944/3410894/catboost/rust-package/catboost-sys/../../../build/build_native.py", line 377, in get_msvs_dir
      raise RuntimeError(f'Microsoft Visual Studio {msvs_version} installation not found')
  RuntimeError: Microsoft Visual Studio 2019 installation not found
  thread 'main' panicked at /Users/jnovikov/.cargo/git/checkouts/catboost-113273d315325944/3410894/catboost/rust-package/catboost-sys/build.rs:38:9:
  Building with build_native.py failed
  ```

This seems to be caused by this commit:
https://github.com/catboost/catboost/commit/14439f2db9eb830a3deb9923ee73f296a14a526b#diff-e9f0de4c0016d609bf5139a6949866b3d80a4cd1b8a03541560f68228842628bR493

jnovikov:
Fixed in https://github.com/catboost/catboost/commit/2015851e55d5aa80b2cbc1e36df9f5c2289a83bc

PR #2581, opened 2024-01-25
Get MSVS directory only if building on Windows

jnovikov:
I hereby agree to the terms of the CLA available at: [https://yandex.ru/legal/cla/?lang=en](https://yandex.ru/legal/cla/?lang=en).

This should solve #2580.


andrey-khropov:
Thank you!

PR #2622, opened 2024-04-02
Fixed bloating temp directory with copies of native libraries on Windows

DKARAGODIN:

1. Read [instructions for contributors](https://catboost.ai/docs/concepts/development-and-contributions.html).
2. Make sure [the code builds](https://catboost.ai/en/docs/concepts/build-from-source).
3. If you add new functionality [add tests](https://catboost.ai/en/docs/concepts/development-and-contributions#run-tests) to check it.
4. [Run existing tests](https://catboost.ai/en/docs/concepts/development-and-contributions#run-tests) to make sure you haven't broken anything.
5. If you haven't already, sign [the Contributor License Agreement](https://yandex.ru/legal/cla/).

Problem: when using catboost in Java a native library has to be loaded into JVM. To do so we extract corresponding native library into temp folder. After JVM shutdown this file stays there. In case when temp folder is system temp folder it may not be a problem. However when catboost used in managed environment like Application Server or Servlet Container, the temp directory will be the temp directory of this environment. Without any cleaning mechanism space on hard drive will be quickly bloated.

In this PR I suggest mechanism that deletes copies of native library in temp directory. To accomplish it I did the following changes:
1. Native libraries copies to temp directory with version.
2. Creating in temp directory empty file with library filename + .lck suffix
3. Before loading library scan temp directory. If we find file that has name and version that we about to load && corresponding .lck file does not exist then delete such file.

This algorithm will work every time JVM shutsdown gracefully. In case of sudden shutdown no remedy exists.

Why this algorithm will work? (Here is my explanation, but I am not 100% sure on it)
The problem is in the fact that native library that we copied to temp folder won't be deleted despite the fact that method deleteOnExit called. That happens because this file was loaded by System classloader and will be released only after System ClassLoader garbage collected. But GC of System ClassLoader means that JVM shut down.
But this won't happen to simple file which is .lck file.

Another approach to solve this problem is to load native library with other ClassLoader. I don't think this is a good idea.

The source of inspiration for this PR can be found here:  https://github.com/xerial/sqlite-jdbc/blob/master/src/main/java/org/sqlite/SQLiteJDBCLoader.java

In the attachment you can find two projects. JNA-WAR and JNA-JAR. Their intention is to reproduce the problem and check the solution. There are dependencies on catboost-prediction and sqllite. JNA-WAR should be run in Servlet Container. I tested on tomcat 10.
[jna_jar.zip](https://github.com/catboost/catboost/files/14837958/jna_jar.zip)
[jna_war.zip](https://github.com/catboost/catboost/files/14837959/jna_war.zip)

### DIFF
--- a/catboost/jvm-packages/common/src/main/java/ai/catboost/common/NativeLib.java
+++ b/catboost/jvm-packages/common/src/main/java/ai/catboost/common/NativeLib.java
@@ -5,18 +5,26 @@ import org.slf4j.LoggerFactory;
 
 import javax.validation.constraints.NotNull;
 import java.io.*;
+import java.net.URL;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Properties;
+import java.util.stream.Stream;
 
 
 public class NativeLib {
     private static final Logger logger = LoggerFactory.getLogger(NativeLib.class);
+    private static final String LOCK_EXT = ".lck";
 
     /**
-     * Load libName, first will try try to load libName from default location then will try to load library from JAR.
+     * Load libName, first will try to load libName from default location then will try to load library from JAR.
      *
      * @param libName
      * @throws IOException
      */
-    public static void smartLoad(final @NotNull String libName) throws IOException {
+    public static synchronized void smartLoad(final @NotNull String libName) throws IOException {
+        cleanup(libName);
         try {
             loadNativeLibraryFromJar(libName);
         } catch (IOException ioe) {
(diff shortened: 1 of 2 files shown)


DKARAGODIN:
I hereby agree to the terms of the CLA available at: [https://yandex.ru/legal/cla/?lang=en].

andrey-khropov:
Related issue in sqlite-jdbc: https://github.com/xerial//issues/80

DKARAGODIN:
Rebased commit on updated master version.

Please, delete Windows tag for this PR. Issue reproduces on Linux as well. You are misled by issue in SQL library reference.

andrey-khropov:
Merged in [f82f4d6](https://github.com/catboost/catboost/commit/f82f4d68a63679f9f6ac380354c1060df40e474e)

andrey-khropov:
> Issue reproduces on Linux as well.

I haven't been able to reproduce it on Linux and macOS. Explanation: https://github.com/xerial/sqlite-jdbc/issues/80#issuecomment-230541695

Issue #2415, opened 2023-06-22
Split descriptions are of insufficient precision


BenModulusLabs:
Problem: Split descriptions are of insufficient precision (see below)
catboost version: 1.2
Operating System: Mac OS X 13.3.1
CPU: Apple Silicon M2
GPU: N/A

The splits used by Catboost are single precision floats, but the split descriptions available via the Python interface use only 6 significant digits.  This level of precision is insufficient to fully specify the 32-bit float (8 significant digits are required).  The result is that distinct but very close together split thresholds appear to identical, and the split information can not be used to trace paths through the decision tree.

(The workaround I've found is to export the model to JSON, and fetch the splits from there, but this seems not ideal).

Example:
```python
model = CatBoostRegressor().load_model('mymodel.h5')
tree_idx = 0
splits = model._get_tree_splits(tree_idx, pool=None)
splits[:5]
```

Outputs:
```
['27, bin=0.914953',
 '32, bin=-0.76942',
 '6, bin=-2.20777',
 '32, bin=-2.43341',
 '13, bin=-3.83913']
```

mihail-shusharin:
Hello, function _get_tree_splits is used only for tree visualisation and should't return 8 significant digits. Your workaround seems to be the best way to fetch splits to use them somewhere else.

BenModulusLabs:
Yes, I agree in principle, but as it currently stands the JSON export is insufficient to reconstruct the tree because of this issue https://github.com/catboost/catboost/issues/2414. Presently it is necessary to use both `_get_tree_splits` and the JSON export together!

It would moreover be more convenient to be able to retrieve the tree information from the Catboost model object directly, without having to write the JSON to disk and read it back in again.  So perhaps the solution is to create a new public method in the Python API that returns the split information (and ideally not as a string)?

PR #2721, opened 2024-08-16
#2415 Fix precision issue in tree split descriptions

Shashank1202:
- Created a new method `get_full_precision_tree_splits` in `core.py` to retrieve tree split descriptions with 8 significant digits.

This change addresses issue #2415, improving the precision of split descriptions in tree visualizations.

### DIFF
--- a/catboost/python-package/catboost/core.py
+++ b/catboost/python-package/catboost/core.py
@@ -7288,3 +7288,17 @@ class _TrainCallbacksWrapper(object):
             if not cb.after_iteration(info):
                 return False
         return True
+    
+### Function to add full Precision split
+def get_full_precision_splits(self, tree_idx, pool= None):
+    """
+    Returns the split information for the specified tree with full precision (8 significant digits).
+
+    :param tree_idx: The index of the tree for which to retrieve the splits.
+    :param pool: The data pool, optional.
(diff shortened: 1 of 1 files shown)


Shashank1202:
Hey can anyone guide me where to add test cases, I went through the document , but couldn't make it.

Your guidance would be highly appreciated!!!

catboost/python-package/catboost/core.py
```diff
@@ -7288,3 +7288,17 @@ def after_iteration(self, info):
             if not cb.after_iteration(info):
                 return False
         return True
+    
+### Function to add full Precision split
+def get_full_precision_splits(self, tree_idx, pool= None):
```
a-holm, line 7293: The new function `get_full_precision_splits` is defined as a standalone function in `core.py`, but it tries to access `self._object`. Since it's not a method of a class, `self` will be undefined, causing a `NameError` when the function is called. Functions operating on the model object should typically be methods within the `CatBoost` class (or its base class) to have access to `self`.

Possibly just an indentation issue

a-holm:
Just a note. The PR description mentions creating a method named `get_full_precision_tree_splits`, but the implemented function is named `get_full_precision_splits`.

Issue #1767, opened 2021-07-12
RMSPE Metric


jonimatix:
Problem:
Is it possible to have the RMSPE (root mean squared percentage error) added to the list of metrics, or someone can create a custom metric for this, would be really appreciated?

Thank you

andrey-khropov:
Implemented in #2741 thanks to @ivan339339 (only on CPU though). Will be included in the next release.

PR #2741, opened 2024-09-21
feat: add RMSPE metric and loss

ivan339339:
Based on [catboost/issues/1767](https://github.com/catboost/catboost/issues/1767).

Added:
- RMSPE function as eval_metric and loss function
- boost_from_average support 
- tests for new functionality

New loss is compatible with Newton `leaf_estimation_method`.


andrey-khropov (changes requested):
ELossFunction::RMSPE has to be added to the list in [ValidateCtrTargetBinarization](https://github.com/catboost/catboost/blob/3da23355716b36bb070f0470de41fe1b08db3fd0/catboost/private/libs/options/catboost_options.cpp#L480)

ivan339339:
All fixes done

catboost/libs/metrics/optimal_const_for_loss.h
```diff
@@ -153,6 +153,32 @@ namespace NCB {
...
+    // based on reasoning from https://gist.github.com/andrey-khropov/b1435bad6e105c54da50d98a734cb3d5
+    inline float CalculateOptimalConstApproxForRMSPE(
+        TConstArrayRef<float> target,
+        TConstArrayRef<float> weights
+    ) {
+        if (target.empty()) {
+            return 0.0;
+        }
+
+        Y_ASSERT(target.size() == weights.size());
```
andrey-khropov, line 165: This is wrong. `weights` can be empty. You should replace this check with 

```c++
Y_ASSERT(weights.empty() || (target.size() == weights.size()));
```

catboost/libs/metrics/optimal_const_for_loss.h
```diff
@@ -153,6 +153,32 @@ namespace NCB {
...
+
+        TVector<float> weightsWithTarget = weights.empty()
```
andrey-khropov, line 167: Please add a comment similar to the comment here: https://github.com/catboost/catboost/blob/70a9b6a086a7dbf67102e50cbfed02b6a0162712/catboost/libs/metrics/optimal_const_for_loss.h#L111

```
// TODO: replace with dispatch by weights.empty()
```

This code makes an unnecessary extra `weightsWithTarget` vector, it should be optimized. But it is ok for now because it is called only once at the start of the training.

andrey-khropov (changes requested):

catboost/libs/metrics/optimal_const_for_loss.h
ivan339339: I replaced it with `float weight = weights.empty() ? 1.0f : weights[idx];` in loop

ivan339339: Or you meant dispatch like this?

```c++
const auto impl = [=] (auto hasWeight) {...}
return DispatchGenericLambda(impl, !weights.empty());
```

andrey-khropov: > ```c++
> const auto impl = [=] (auto hasWeight) {...}
> return DispatchGenericLambda(impl, !weights.empty());
> ```

Yes. But I don't consider this change to be necessary to merge this PR.

ivan339339: Added `TODO` comment message

andrey-khropov (approved):

Issue #2744, opened 2024-09-24
Apache Spark PySpark application is not stopping


KernelA:
Problem: 
```python

with SparkSession.builder.appName("TestCatboost").config(
    "spark.sql.parser.ansi.enabled", "true"
).getOrCreate() as spark:
    import catboost_spark  # noqa: E402
    ...
    trainPool.quantize(
            catboost_spark.QuantizationParams(
                borderCount=30,
                featureBorderType=catboost_spark.EBorderSelectionType.GreedyLogSum,
            )
        )
    reg = (
            catboost_spark.CatBoostRegressor(iterations=50)
            .setTrainDir("catboost_train")
            .setLossFunction("PairLogitPairwise")
        )
    model = reg.fit(trainPool)
```

```
spark-submit --master "local[4]" \
    --jars jars/catboost-common-1.2.7.jar,jars/catboost-spark-macros_2.12-1.2.7.jar,jars/catboost-spark_3.3_2.12-1.2.7.jar,jars/classgraph-4.8.98.jar \
    --py-files jars/catboost-common-1.2.7.jar,jars/catboost-spark-macros_2.12-1.2.7.jar,jars/catboost-spark_3.3_2.12-1.2.7.jar,jars/classgraph-4.8.98.jar \
    --conf spark.driver.memory="2G" \
    --conf spark.task.cpus=4 \
    ....
```

Logs:
```
48:	learn: 0.0458000	total: 4s	remaining: 81.6ms
49:	learn: 0.0444066	total: 4.08s	remaining: 0us
QueryFullTime: 3.724877
QueryExecutionTime: 1.198499
Skipping test eval output
0.07039295165 min passed
2024-09-24 23:40:43,052 - INFO - py4j.clientserver - Closing down clientserver connection
```
catboost version: catboost-spark_3.3_2.12-1.2.7
Operating System: macOS 13.4
CPU: 2,6 GHz 6-Core Intel Core i7
GPU: no
PySpark: 3.3.0
Python: 3.10
Java: 11

In local and cluster modes, the same issue occurs as:
#2151

Issue #2151, opened 2022-08-21
【Question】catboost for spark how to exit 


chencjiajy:
Problem: catboost for spark application not exit after finished, it hangs....., `spark.stop()` not work。 how to make the application exit after training. （ps:both IDEA and spark-submit not exit)
catboost version:1.0.6

[example](https://github.com/catboost/catboost/blob/313d5a014b000933feecd29f214220738cf64d85/catboost/spark/catboost4j-spark/core/src/test/scala/ai/catboost/spark/CatBoostRegressorTest.scala#L1474)

<img width="1123" alt="how to exit catboost for spark" src="https://user-images.githubusercontent.com/19321967/185774812-7b64a626-ca6b-4f4c-8590-c1d881cd2719.png">

shuchih:
same problem here

tqtensor:
We have the same problem.

art777nn:
We have the same problem.

Tried to run different versions catboost-spark:
catboost-spark_3.4_2.13:1.2.2, 1.2, 1.2.1
catboost-spark_3.4_2.12:1.2.2, 1.2, 1.2.1

Context:
Kubernets 1.27.7 
Spark 3.4.2
Scala 2.13 or 2.12
Catboost 1.2.2
PySpark 3.4.2

Driver does not stop after fit call.

Attached extended logs that are an output after the python script is completed. You can see that at the end of logs PoolingHttpClientConnectionManager attempts to close idle connections infinitely.

Log line output:
`PoolingHttpClientConnectionManager: Closing connections idle longer than 60000 MILLISECONDS
`
[log_catboost.txt](https://github.com/catboost/catboost/files/14049389/log_catboost.txt)

andrey-khropov:
Fixed in https://github.com/catboost/catboost/commit/2cdb8a2d52777538f3f07d8692132abd99dc18e2, the good news is that is was just auxiliary threads not shut down at the end of the program execution, the main computation logic had not been affected, the training, model prediction etc. results should be valid.

art777nn:
> Fixed in [2cdb8a2](https://github.com/catboost/catboost/commit/2cdb8a2d52777538f3f07d8692132abd99dc18e2), the good news is that is was just auxiliary threads not shut down at the end of the program execution, the main computation logic had not been affected, the training results should be valid.

Thank you! This commit cured the application! Can you please tell me when we can expect a patch in maven?

andrey-khropov:
[...]

PR #2779, opened 2024-11-23
Fix workers hanging

Shamann:
This PR aims to fix the issue of hanging catboost workers leading to spark applications can't exit properly (#2151, #2744).

The root cause is that the internal CatBoostWorker thread pool is never shutdown.

The unit tests are unable to catch this kind of issues, since the maven surefire plugin shuts forked JVM down, so all non daemon threads die. A junit verifier is added to check for non daemon threads that are still running at the end of the test case.

### DIFF
--- a/catboost/spark/catboost4j-spark/core/src/main/scala/ai/catboost/spark/impl/Workers.scala
+++ b/catboost/spark/catboost4j-spark/core/src/main/scala/ai/catboost/spark/impl/Workers.scala
@@ -94,7 +94,8 @@ private[spark] class CatBoostWorker(partitionId : Int) extends Logging {
 
       val workerListeningPort = if (workerListeningPortParam != 0) { workerListeningPortParam } else { TrainingDriver.getWorkerPort() }
 
-      val ecs = new ExecutorCompletionService[Unit](Executors.newFixedThreadPool(2))
+      val service = Executors.newFixedThreadPool(2)
+      val ecs = new ExecutorCompletionService[Unit](service)
 
       val partitionId = this.partitionId
       val sendWorkerInfoFuture = ecs.submit(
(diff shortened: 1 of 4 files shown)


catboost/spark/catboost4j-spark/core/src/main/scala/ai/catboost/spark/impl/Workers.scala
```diff
@@ -94,7 +94,8 @@ private[spark] class CatBoostWorker(partitionId : Int) extends Logging {
 
       val workerListeningPort = if (workerListeningPortParam != 0) { workerListeningPortParam } else { TrainingDriver.getWorkerPort() }
 
-      val ecs = new ExecutorCompletionService[Unit](Executors.newFixedThreadPool(2))
+      val service = Executors.newFixedThreadPool(2)
```
andrey-khropov, line 97: I'd call it `ecsPool` for uniformity with `ecsPool` is CatBoostPredictor.scala.

catboost/spark/catboost4j-spark/core/src/main/scala/ai/catboost/spark/impl/Workers.scala
```diff
@@ -126,6 +127,8 @@ private[spark] class CatBoostWorker(partitionId : Int) extends Logging {
         ()
       )
 
+      service.shutdown()
```
andrey-khropov, line 130: - That's too early, the futures might not even have started to execute.
- In case of exception in the code of futures creation and submission this code won't be called.

It's appropriate to move the `try {` block opening below higher just after the creation of `service` / `ecsPool` and put `service`/`ecsPool`.shutdown() in the corresponding `finally` block, similar to how it is done in CatBoostPredictor.scala.

andrey-khropov (changes requested):

andrey-khropov (approved):

andrey-khropov:
This PR has been merged. Thank you.

PR #2791, opened 2024-12-30
fix a typo in core.py

Creative-Ataraxia:
Bernoulli is mistyped as Bernoully on line 1938 in core.py

PR #2792, opened 2024-12-30
fix a typo in core.py

Creative-Ataraxia:
Bernoulli is mistyped as Bernoully on line 1939 in core.py


andrey-khropov:
Duplicate of #2791

Issue #2491, opened 2023-09-11
[R] Installation error on mac-arm: some hard-coded temporary paths could not be fixed


david-cortes:
When trying to install `catboost==1.2.1` from the files in the github release tag on a mac-arm system with `R CMD INSTALL`, it throws an error saying
> ERROR: some hard-coded temporary paths could not be fixed
This happens with both `catboost-R-Darwin-1.2.1.tgz` and `catboost-R-darwin-universal2-1.2.1.tgz`.

It seems the error has been there for a while as it also happens with `catboost==1.2`.

The linux files work correctly though, as do the python releases in PyPI.

andrey-khropov:
This is a known issue that can be remedied by adding `--no-staged-install` to the installation options.

andrey-khropov:
I've decided to reopen this issue. I can be worked around by disabling staged installation and it is documented now but it is not ideal because it requires additional troubleshooting by users during installation and the benefits of staged installation are lost on macOS Arm platform.

So the underlying issues that cause this error should be investigated more thoroughly.

PR #2809, opened 2025-01-27
Fix R build on MacOS arm64

jeroen:
This fixes the problem that causes the package to fail installation on arm64 MacOS machines:

```
   ===== ./catboost.Rcheck/00install.out =====
  * installing *source* package ‘catboost’ ...
  ** using staged installation
  checking for R_HOME... /Library/Frameworks/R.framework/Resources
  checking for R... /Library/Frameworks/R.framework/Resources/bin/R
  *** creating .so -> .dylib symlink for working with different R installations
  checking for local CATBOOST_DYNLIB... no
  checking whether we can fetch CatBoost dynlib... downloading CatBoost (libcatboostr.dylib - v1.2.7)
  trying URL 'https://github.com/catboost/catboost/releases/download/v1.2.7/libcatboostr-darwin-universal2-v1.2.7.dylib'
  Content type 'application/octet-stream' length 62835520 bytes (59.9 MB)
  ==================================================
  downloaded 59.9 MB
  
  CatBoost fetch successful
  configure: creating ./config.status
  ** libs
  make: Nothing to be done for `all'.
  installing to /Users/runner/work/mlr-org/mlr-org/catboost.Rcheck/00LOCK-catboost/00new/catboost/libs
  ** R
  ** inst
  ** byte-compile and prepare package for lazy loading
  ** help
  *** installing help indices
  ** building package indices
  ** testing if installed package can be loaded from temporary location
  ** checking absolute paths in shared objects and dynamic libraries
  ERROR: some hard-coded temporary paths could not be fixed
  * removing ‘/Users/runner/work/mlr-org/mlr-org/catboost.Rcheck/catboost’
  ```

A better solution is to use a static library for libcatboostr instead of a dynamic one, but in the mean while this works.

### DIFF
--- a/catboost/R-package/DESCRIPTION
+++ b/catboost/R-package/DESCRIPTION
@@ -54,3 +54,4 @@ BugReports: https://github.com/catboost/catboost/issues
 Suggests: caret, testthat, tibble, e1071
 Biarch: FALSE
 Encoding: UTF-8
+StagedInstall: FALSE


jeroen:
Can we merge this? 🙏

andrey-khropov:
> Can we merge this? 🙏

I've checked using CatBoost release 1.2.7 and this error occurs only on macOS on Arm64 platform.
So, I'd rather not disable staged installation globally on all platforms because it provides real benefits.

It is [mentioned in the documentation that this problem can be fixed be disabling staged installation if necessary](https://catboost.ai/docs/en/installation/r-installation-binary-installation#troubleshooting).

But I agree that it is not ideal so I've decided to reopen #2491.

jeroen:
Maybe you can merge this solution until you fix https://github.com/catboost/catboost/issues/2491 ? Currently the package is uninstallable for most macos users.

andrey-khropov:
> Currently the package is uninstallable for most macos users.

Why? Just add "--no-staged-install" option to `INSTALL_opts` parameter as described in documentation.

jeroen:
I am the maintainer of https://r-universe.dev and we build 26000+ R packages. It is not realistic to maintain custom install flags for each of those packages. The same holds for other R packages repositories. So it would be really nice if the package can be built with the default flags.

andrey-khropov:
> Another solution is that you could add a line to your package `/configure.ac` that automatically writes `StagedInstall: FALSE` to the DESCRIPTION file on the platforms where you require this, instead of expecting the user to figure this out

Thank you, that's an interesting suggestion. I'll take a look.

PR #2835, opened 2025-03-21
Optimize Lossguide

Levachev:
Optimize Lossguide with subtract trick

PR #2883, opened 2025-05-28
Optimize Lossguide

Levachev:
Optimizing Lossguide with the subtract trick.

When testing, it was found that the result will be different(compared to the master version) if you do not set the parameters of CatBoostRegressor
random_strength=0,
bootstrap_type="No",
has_time=True
Otherwise, the result will be different due to the fact that Ctx->LearnProgress->Rand.GenRand() will be different when splitting the leaf


Evgueni-Petrov-aka-espetrov:
copy of https://github.com/catboost/catboost/pull/2835 which cannot be merged automatically because of sync engine update

Evgueni-Petrov-aka-espetrov:
Shipped!

Levachev:
I hereby agree to the terms of the CLA available at: [link](https://yandex.ru/legal/cla/ru/?lang=ru).

Evgueni-Petrov-aka-espetrov:
Shipped!

Evgueni-Petrov-aka-espetrov:
Shipped!

Issue #2961, opened 2025-11-01
MultiRMSEWithMissingValues for GPU implementation


mkotyushev:
Hello,

I have implemented the `MultiRMSEWithMissingValues` loss for GPU for my side project needs, I would like to contribute it to the main repository. I added the code of CUDA-kernels, registered them, and added the `else if` cases and enums in ~ same way as for `MultiRMSE` loss. Now it compiles & could be installed via wheel & used from python.

The remaining work from my perspective is to:
- provide `TargetBinarization` params in the right way, as I hardcoded the `NanMode` and `BorderCount` for now,
- run the current tests & add tests for GPU implementation, as I only compared GPU vs CPU loss over iterations on a random dataset (it converges) and used the loss in my project without any obvious issues,
- check any other places where the new loss or binarization params for it should be registered / added.

I could open a draft PR with the current state & do the remaining work with your guidance if it sounds good to you.

Thanks a lot!

andrey-khropov:
Yes, of course. That would be great!

PR #2962, opened 2025-11-02
Implement MultiRMSEWithMissingValues on GPU

mkotyushev:
Following #2961 
I open this as PR and not draft PR as I need some guidance from the maintainers.

DONE: 1. Read [instructions for contributors](https://catboost.ai/docs/concepts/development-and-contributions.html).
DONE: 2. Make sure [the code builds](https://catboost.ai/en/docs/concepts/build-from-source).
TODO: 3. If you add new functionality [add tests](https://catboost.ai/en/docs/concepts/development-and-contributions#run-tests) to check it.
DONE: 4. [Run existing tests](https://catboost.ai/en/docs/concepts/development-and-contributions#run-tests) to make sure you haven't broken anything.

# PR structure

pr_struct_annot

# Running existing tests

The same two test fail in both `master` and `multi-rmse-with-missing-values-gpu` branches in my runs.

`master` commit `3335a0`

    ================================================================================ short test summary info =================================================================================
    FAILED test.py::test_overfit_detector_with_resume_from_snapshot_and_metric_period[Iter-Ordered] - assert 23 == 8
    FAILED test.py::test_overfit_detector_with_resume_from_snapshot_and_metric_period[IncToDec-Ordered] - assert 79 == 8
    ==================================================== 2 failed, 1052 passed, 27 skipped, 10 xfailed, 27 warnings in 475.53s (0:07:55) =====================================================

`multi-rmse-with-missing-values-gpu` commit `232957`

    ================================================================================ short test summary info =================================================================================
    FAILED test.py::test_overfit_detector_with_resume_from_snapshot_and_metric_period[Iter-Ordered] - assert 23 == 8
    FAILED test.py::test_overfit_detector_with_resume_from_snapshot_and_metric_period[IncToDec-Ordered] - assert 79 == 8
    ==================================================== 2 failed, 1052 passed, 27 skipped, 10 xfailed, 27 warnings in 467.77s (0:07:47) =====================================================

# Compare with the reference CPU implementation

I compared CPU vs GPU for the proposed implementation and for the `MultiRMSE`, training 5 `random_seed` model params and a few rounds of randomly generated data, below are plots for one data seed (other data seeds looks similar). Here, the eval data = train data.

<p align="center">
  multirmsewithmissingvalues_cpu_vs_gpu
  rmse_cpu_vs_gpu
  <br>
  <sub>MultiRMSEWithMissingValues: CPU vs GPU</sub>
  &nbsp;&nbsp;&nbsp;&nbsp;
  <sub>MultiRMSE: CPU vs GPU</sub>
</p>

I expect some difference along the training process but always convergence to the same value at the end, and it seems to hold true, although on toy data & eval = train.

# TODO

1. I think setting the `NanMode` to `ENanMode::Min` is ok, but I don't know the difference between min and max. And I don't really understand how to set `BorderCount` and whether we should do that in `catboost/cuda/train_lib/train.cpp` or there is a better place for it. 

E.g. 

- in `catboost/python-package/catboost/_catboost.pyx` as for `quantizedFeaturesInfo`
- or in same `train.cpp` but in `UpdateGpuSpecificDefaults` function, moving it before `TBinarizedFeaturesManager` creation

2. Also, there might be other places where `TBordersBuilder::operator()` or `TGridBuilderBase` methods are called (e.g. in `catboost/cuda/gpu_data/gpu_binarization_helpers.cpp`) which I didn't hit running training from python, for them `NanMode` should be provided (default is `ENanMode::Forbidden`).

3. Finally, unit test should be added, similar to `test_multiregression_with_missing_values` in `catboost/pytest/test.py` comparing metric across iterations to the canonical values. And ideally comparing the final values between CPU and GPU as in the tests above.

I would happily address those issues with a bit of a guidance from you. (1) and (2) are more or less clear, we just need to figure out where to put the code, as for (3): is there any guide on how to add such tests?


Evgueni-Petrov-aka-espetrov:
thank you for your pr!

could you please add tests to catboost/cuda/targets/ut, catboost/pytest and catboost/python-packages/ut/medium similar to the following tests?

or please reply with a link to the tests, if they are somewhere in the pr

https://github.com/catboost/catboost/blob/673e34746e11d6ebdea899c3aa487b9a43f112d7/catboost/pytest/cuda_tests/test_gpu.py#L1761

https://github.com/catboost/catboost/blob/673e34746e11d6ebdea899c3aa487b9a43f112d7/catboost/python-package/ut/medium/test.py#L375

to run it on gpu, we need to add task_type parameter, like in test below
https://github.com/catboost/catboost/blob/673e34746e11d6ebdea899c3aa487b9a43f112d7/catboost/python-package/ut/medium/test.py#L4473

i am sorry for taking time to respond.

Evgueni-Petrov-aka-espetrov:
i think it makes sense to compare predictions of (a model trained with) your implementation against predictions of the cpu version on some small random dataset with some reasonable tolerance

mkotyushev:
@Evgueni-Petrov-aka-espetrov thanks for the review! I've pushed an updated version. Could you please approve the workflow? CI is green on [my repo fork](https://github.com/mkotyushev/catboost/pull/1), and I ran the `catboost/pytest/cuda_tests` GPU suite locally (576 passed). All 3 TODOs and both your comments (adding tests and prediction comparison) are resolved.

Changes since the previous version:

1. Rebased onto the `master` (329bc3).
2. Resolved TODO 1 and 2: NaNs are removed before building the borders now, as `BuildTargetClassifier` does on CPU via `BestSplit(..., filterNans=true)`, so no border builder changes are necessary and `grid_creator.*` changes are reverted. Thus, no mode switching (TODO 1) / code changes and no other places using `TBordersBuilder::operator()` are affected (TODO 2).
3. Fix 2 bugs in the PR code: 
   - The second-derivative kernel wrote the weight into every target dimension, but the leaves estimator (`TBinOptimizedOracle::WriteSecondDerivatives`) allocates the row buffer with only `row + 1` columns, so it wrote out of bounds. It now mirrors `MultiRMSESecondDerImpl` and writes only the diagonal element of the requested row, masked by the missing-value check. (`Target` was also missing from the kernel's `Y_SAVELOAD_DEFINE`.)
   - The metric reused `MultiRMSE`'s single `(SSE, weight)` pair, but `TMultiRMSEWithMissingValues` on CPU keeps a `(SSE, weight)` pair per target dimension and returns `sqrt(sum_d SSE_d / W_d)` — so the reported GPU metric diverged from CPU whenever a target was missing, while the training was ok. Added a `MultiRMSEWithMissingValuesStats` kernel producing the per-dimension pairs.
4. Added tests and GPU-vs-CPU comparison:
    - `catboost/pytest/cuda_tests/test_gpu.py` — `test_multirmse_with_missing_values`, `..._with_cat_features`, and `..._with_many_missing_values`, comparing CPU vs GPU predictions with `rtol=atol=1e-6` like `test_multirmse`. The last uses a generated dataset with a 30% missing rate (the `multiregression_with_missing` dataset has only two missing values) and also compares the metric written to the test-error log.
    - `catboost/python-package/ut/medium/test.py` — `test_multiregression_with_missing_values`, parameterized by `task_type`. CPU canonical data is included; the GPU canonical data needs to be produced on your reference GPU.
    - `catboost/cuda/targets/ut/test_multi_rmse_with_missing_values.cpp` — value, first/second derivatives and metric stats vs a CPU reference over several document counts, target dimensions and missing rates; also asserts the second-derivative kernel writes nothing above the diagonal of the requested row.
5. Prediction difference check with tolerance: On a 500x5 dataset, 3 targets, 30% missing, `border_count` pinned to 254 (default differs for CPU and GPU) on both sides, `max|cpu - gpu|` is 3.2e-07 at 100 iterations and 4.0e-08 at 10 — the same order as plain `MultiRMSE` (3.7e-07), so the missing-value handling adds no divergence of its own.

Happy to adjust anything.

P.S. I used Claude Code for those later changes, it is reflected in the commit messages. I reviewed the full PR diff later.

mkotyushev:
@andrey-khropov Hi, could you please approve the workflow?

PR #2968, opened 2025-11-13
Automatically disable staged install on Mac arm64

### LINKED ISSUE #2491: [R] Installation error on mac-arm: some hard-coded temporary paths could not be fixed

jeroen:
Fixes #2491

Automatically sets the correct flags on MacOS such that users and build servers don't have special-case building of this package by setting `--no-staged-install`. See discussion at https://github.com/catboost/catboost/pull/2809

### DIFF
--- a/.github/workflows/test_per_platform.yaml
+++ b/.github/workflows/test_per_platform.yaml
@@ -315,6 +315,6 @@ jobs:
         name: Check R package
         shell: pwsh
         run: |
-          Rscript -e "install.packages('remotes', Ncpus = parallel::detectCores()); remotes::install_local('catboost_R_package/catboost-R-darwin-universal2.tgz', INSTALL_opts = c('--no-multiarch', '--no-staged-install', '--install-tests'))"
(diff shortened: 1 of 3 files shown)


topepo:
One reason this is a good idea is that we have a single, clean way to install the R package using GitHub Actions. We test across a wide range of OS and R version combinations, and writing different install code for one OS isn't ideal.

andrey-khropov:
Unfortunately, it looks like this approach does not work.

I see the same error

`ERROR: some hard-coded temporary paths could not be fixed`

if I remove `--no-staged-install` in CI.

You can see it in  'Check R package' stage here: https://github.com/catboost/catboost/actions/runs/19391087947/job/55492037883?pr=2968

CI run shows a green status but it is in fact incorrect (seems like return codes are not checked for some commands, it will be dealt with in a separate issue #2969 ).

PR #2973, opened 2025-11-28
[rust-package] PR: Add feature "use_prebuilt"

marcomq:
- adding optional feature "use_prebuilt" to avoid compiling the C code all the time. Default is status quo - use_prebuilt is off

### DIFF
--- a/catboost/rust-package/Cargo.lock
+++ b/catboost/rust-package/Cargo.lock
@@ -73,7 +73,6 @@ version = "0.1.1"
 dependencies = [
  "approx",
  "catboost-sys",
- "libc",
 ]
 
 [[package]]


marcomq:
Ok. Just added 
https://github.com/catboost/catboost/pull/2975
for the functional changes. Keeping this here for optionally using prebuilt catboost instead of always building it.

alexeysofin:
I think there's one more case worth covering, when a custom built library must be used, e.g. built with custom flags or just outside of rust package for caching.

This needs neither downloading prebuilt binaries nor building it in build.rs, just compiling the bindings and using whatever is supplied with LD_LIBRARY_PATH or LD_PRELOAD in runtime

andrey-khropov:
I've researched this issue and it looks like specifying the way to use `catboostmodel` library by Cargo features could be problematic.
Cargo features are boolean flags, meant to be additive and designed to add optional functionality without changing anything for packages that do not enable these features.

For example, if you have a package `X` with features `f1` and `f2`,  package `Y` depends on `X` with feature `f1` and package `Z` depends on `X` with feature `f2` then Cargo will merge features demanded by `Y` and `Z` from `X`, i.e. `X` will be built with features `f1` and `f2`.

But the nature of the choice whether to build `catboostmodel` from source or download the stable release is mutually exclusive. It cannot be toggled by the `use_prebuilt` feature as proposed in this PR because if a single package that depends on `catboost` or `catboost-sys` enables this feature it will be enabled for all the packages that depend on `catboost` or `catboost-sys`.

I think the right design is to have a switch with three possible options to use `catboostmodel` library:

- Build from source (I think it should be kept as the default to maintain backward compatibility)
- Use prebuilt binaries from the release on GitHub
- Use from system (when you already have it in your system)

And I think the right way to implement such a switch would be an environment variable.

I propose to have an environment variable called `CARGO_CATBOOSTMODEL_LIB_SOURCE` with the possible values:

• `BUILD` - Build from source (I think it should be kept as the default to maintain backward compatibility)
• `BINARY_RELEASE` - Use prebuilt binaries from the release on GitHub
• `SYSTEM` - Use from system (when you already have it in your system)

If it is not defined the default behavior will be the same as `BUILD`.

I also researched other well-known libraries that have similar issues (like `ort-sys`, `rusty_v8`, `rust-xgboost`, `tensorflow-sys`, `torch-sys`). They use either Cargo features or environment variables but unfortunately there are no commonly accepted guidelines in Rust ecosystem how to implement this, each library does it in its own slightly different way.

P.S. @marcomq, I can do the necessary modifications myself because I am allowed to modify the branch for this PR (your authorship for your commits will be preserved) but I want to make sure that you agree if I proceed in such a manner.

marcomq:
> P.S. @marcomq, I can do the necessary modifications myself because I am allowed to modify the branch for this PR (your authorship for your commits will be preserved) but I want to make sure that you agree if I proceed in such a manner.

Sure - go ahead. I currently don't have much time to perform the changes by myself.

PR #3047, opened 2026-02-25
Fix feature name parser misreading hyphens as range operators

### LINKED ISSUE #2983: Having a "-" in the column header name raises an error
Header name = "non-TB"
```
selection_result = model.select_features(train_pool,features_for_select=features,num_features_to_select=features_num,plot=True,train_final_model=True)

MetricVisualizer(layout=Layout(align_self='stretch', height='500px'))
Traceback (most recent call last):
  File "C:\Users\ssd1231231312\Downloads\AI training\test_catboost_feature_all.py", line 35, in <module>
    selection_result = model.select_features(train_pool,features_for_select=features,num_features_to_select=features_num,plot=True,train_final_model=True)
  File "C:\Program Files\Python310\lib\site-packages\catboost\core.py", line 4519, in select_features
    summary = self._object._select_features(train_pool, test_pool, params)
  File "_catboost.pyx", line 5621, in _catboost._CatBoost._select_features
  File "_catboost.pyx", line 5635, in _catboost._CatBoost._select_features
_catboost.CatBoostError: catboost/libs/data/feature_names_converter.h:33: String 'TB' is not a feature name
```

gautamvarmadatla:
Feature names containing hyphens caused a crash in `select_features()` because `TIndicesMapper::Map()` unconditionally split any string that was not a numeric index or `#tag` on `'-'` to parse it as an index range, and the return value of
`TryCollectInto` was not captured. A name like `"non-TB"` splits into exactly two tokens, so the code erroneously attempted a range lookup on the two halves, producing a confusing error that named part of the original feature name as the culprit.

I added an exact `IndicesFromNames` lookup before the range parsing path, so hyphenated names are resolved without ever reaching the range logic. I also now capture `TryCollectInto`'s return value and assert it with `CB_ENSURE`, so
strings that are neither a known feature name nor a valid range produce a clear error instead of silently misinterpreting partial tokens. Added a regression test.

Fixes #2983

### DIFF
--- a/catboost/libs/data/feature_names_converter.h
+++ b/catboost/libs/data/feature_names_converter.h
@@ -67,9 +67,14 @@ namespace {
                 CB_ENSURE(tagIndices, TStringBuf() << "There is no tag '#" << tag << "' in pool metainfo");
                 indices->insert(indices->end(), tagIndices->begin(), tagIndices->end());
             } else {
+                if (IndicesFromNames.contains(str)) {
+                    indices->push_back(IndicesFromNames.at(str));
+                    return;
+                }
                 auto left = str;
                 auto right = str;
-                StringSplitter(str).Split('-').TryCollectInto(&left, &right);
+                const bool isRange = StringSplitter(str).Split('-').TryCollectInto(&left, &right);
+                CB_ENSURE(isRange, "String '" + str + "' is not a feature name");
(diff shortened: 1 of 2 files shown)


andrey-khropov (approved):

andrey-khropov:
The failed test on Windows is not related to this PR

PR #3050, opened 2026-02-27
[ENH] Making training directory hidden by default

ThibaultDECO:
Making training directory hidden by default, which is easier when working on a project with multiple folders.
Also fixes some warnings in the tests.

### DIFF
--- a/catboost/python-package/catboost/_catboost.pyx
+++ b/catboost/python-package/catboost/_catboost.pyx
@@ -6219,9 +6219,10 @@ cdef TCustomTrainTestSubsets _make_train_test_subsets(_PoolBase pool, folds) exc
     if hasattr(folds, 'split'):
         if arr_group_ids.Defined():
             flatted_group = _CreateNumpyUI64ArrayView(arr_group_ids.GetRef().data(), arr_group_ids.GetRef().size())
+            folds = folds.split(X=np.zeros(num_data), y=pool.get_label(), groups=flatted_group)
         else:
             flatted_group = np.zeros(num_data, dtype=int)
-        folds = folds.split(X=np.zeros(num_data), y=pool.get_label(), groups=flatted_group)
+            folds = folds.split(X=np.zeros(num_data), y=pool.get_label())
 
     cdef TVector[TVector[ui32]] custom_train_subsets
     cdef TVector[TVector[ui32]] custom_test_subsets
--- a/catboost/python-package/catboost/core.py
+++ b/catboost/python-package/catboost/core.py
@@ -1550,7 +1550,7 @@ def _clear_training_files(train_dir):
 
 
 def _get_train_dir(params):
-    return params.get('train_dir', 'catboost_info')
+    return params.get('train_dir', '.catboost_info')
 
 
 def _get_catboost_widget(train_dirs):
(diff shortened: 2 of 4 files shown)


catboost/python-package/catboost/core.py
```diff
@@ -1550,7 +1550,7 @@ def _clear_training_files(train_dir):
 
 
 def _get_train_dir(params):
-    return params.get('train_dir', 'catboost_info')
+    return params.get('train_dir', '.catboost_info')
```
andrey-khropov, line 1553: I don't think it is a good idea because:
- It is better to have a visible directory that you are aware of instead of it existing silently and wasting system resources
- It will be a breaking change. The current default name `catboost_info` [is documented](https://catboost.ai/docs/en/references/training-parameters/output#train_dir) so some people could rely on this name.
- Dot prefix won't make it hidden on Windows, so the behavior will not be consistent across all platforms.

And I don't quite understand the problem you are trying to solve.

> which is easier when working on a project with multiple folders

What will be easier if the training directory is hidden?

catboost/python-package/catboost/datasets.py
```diff
@@ -19,7 +19,7 @@ def _extract(src_file, dst_dir='.'):
     os.chdir(dst_dir)
     try:
         with tarfile.open(src_file, 'r:gz') as f:
-            f.extractall()
+            f.extractall(filter="data")
```
andrey-khropov, line 22: Why do you want this change?

catboost/python-package/ut/medium/test.py
```diff
@@ -10103,13 +10103,10 @@ def test_feature_tags_interface():
     cat.fit(pool)
     assert np.array_equal(np.where(cat.feature_importances_ == 0)[0], [0, 1, 2, 3, 7])
     cat = CatBoostClassifier()
-    plot_file = test_output_path('plot.html')
     result = cat.select_features(
         pool,
         features_for_select=["#tag1", "#tag2"],
-        num_features_to_select=3,
-        plot=True,
-        plot_file=plot_file
```
andrey-khropov, line 10112: I suspect that you want to avoid displaying the plot, right? 
Testing the plot is valuable, so I propose to turn on/off displaying a plot via a configurable parameter when running tests.

catboost/python-package/catboost/_catboost.pyx
```diff
@@ -6219,9 +6219,10 @@ cdef TCustomTrainTestSubsets _make_train_test_subsets(_PoolBase pool, folds) exc
     if hasattr(folds, 'split'):
         if arr_group_ids.Defined():
             flatted_group = _CreateNumpyUI64ArrayView(arr_group_ids.GetRef().data(), arr_group_ids.GetRef().size())
+            folds = folds.split(X=np.zeros(num_data), y=pool.get_label(), groups=flatted_group)
         else:
             flatted_group = np.zeros(num_data, dtype=int)
-        folds = folds.split(X=np.zeros(num_data), y=pool.get_label(), groups=flatted_group)
+            folds = folds.split(X=np.zeros(num_data), y=pool.get_label())
```
andrey-khropov, line 6225: Why do you want this change?

andrey-khropov (changes requested):
Please, split unrelated changes into different PRs.

PR #3054, opened 2026-03-05
Addition of Piecewise Quantile Loss

### LINKED ISSUE #3053: Feature Contribution: Piecewise Quantile Loss
We, Wayfair’s Supply Chain and Retail Technology Science team, are interested in contributing a new loss function called PiecewiseQuantile loss that we have implemented. It uses different quantile values based on where a sample’s target/ground truth value falls within user-provided intervals. 

As an example:
`PiecewiseQuantile:boundaries=0,1;quantiles=0.25,0.5,0.75`
This assigns: 
- alpha=0.25 when y <= 0
- alpha=0.5 when 0 < y <= 1
- alpha=0.75 when y > 1. 

The length of these input lists are not fixed. To be clear, these quantiles are with respect to the other samples in the respective intervals. 

We have found that such a loss function has worked well for us in various projects over the years, and think it would have value to the larger community. Examples of use cases could include demand forecasting or financial risk modeling. Happy to open a PR if there is interest. 

Thanks!

tatron:
This PR adds a new loss function called **PiecewiseQuantile** and closes #3053 . It uses different quantile values based on where a sample’s target/ground truth value falls within user-provided intervals.

As an example:
`PiecewiseQuantile:boundaries=0,1;quantiles=0.25,0.5,0.75`
This assigns:

- alpha=0.25 when y <= 0
- alpha=0.5 when 0 < y <= 1
- alpha=0.75 when y > 1.

The length of these input lists are not fixed. To be clear, these quantiles are with respect to the other samples in the respective intervals.

We include the loss function, its gradients, and the corresponding metric. We also include unit tests for all of our additions.

### DIFF
--- a/catboost/libs/helpers/quantile.cpp
+++ b/catboost/libs/helpers/quantile.cpp
@@ -1,5 +1,7 @@
 #include "quantile.h"
 
+#include <catboost/private/libs/algo_helpers/quantile_selection.h>
+
 #include <util/generic/algorithm.h>
 #include <util/generic/array_ref.h>
 #include <util/generic/vector.h>
@@ -7,6 +9,7 @@
 #include <util/generic/ymath.h>
 
 #include <algorithm>
+#include <cmath>
 
 namespace {
     struct TValueWithWeight {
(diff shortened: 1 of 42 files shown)


tatron:
I noticed that one of the checks failed, but this seems to be due to a Github internal server issue at the time of the CI run, rather than our code. I can confirm that when I built the wheel and ran tests locally on my MacOS, it passed without issue.

PR #3082, opened 2026-04-17
Add Pool.set_label() for post-construction label mutation

fingoldo:
# Add `Pool.set_label()` for post-construction label mutation

## Summary

Adds `Pool.set_label(y)` — a counterpart to `Pool.set_weight` — that replaces labels on an already-constructed `Pool` in place. Enables Pool reuse across targets and weighting schemes without rebuilding the feature data.

## Motivation

Training multiple models on the same large dataset with different targets currently forces a full Pool reconstruction for every label change. On 100 GB+ data each Pool build takes minutes; across N targets this is pure waste. `set_weight()` already exists for weights — `set_label()` was the missing piece.

## Behavior changes

- `model.fit(pool, y=new_labels)` previously raised; now routes through `pool.set_label(y)`. A `UserWarning` fires when this overwrites existing labels. Same for the three `model.score(pool, y=...)` methods.
- `fit(unlabeled_pool, y=...)` now succeeds (previously raised).
- `fit(pool, eval_set=pool, y=y)` — or any `eval_set` shape that aliases `X` (Pool, `(X, y)` tuple, list of either) — now raises rather than silently mutating the shared Pool and invalidating the hold-out.
- `pool.get_label()` after `set_label` returns the input dtype; internal storage stays `float32`.
- `fit()` emits a `UserWarning` when the training Pool and a Pool-typed `eval_set` have different target dtypes.

Two pre-existing infrastructure issues surfaced through `set_label` usage and are fixed here:
- `SetNumericTarget` rejected any existing non-`Float` target type. Now accepts `Boolean`/`Integer`/`Float`/`None` and rejects only `String`, so `Pool(X, label=np.int64_array).set_label(...)` works.
- `_read_pool` hardcoded the Python-side `target_type` to `str` for every file format, so `Pool("quantized://...").get_label()` returned string-cast numbers. Now detected from the native `RawTargetData.GetTargetType()`.

## Benchmark

100 k × 576 float32, CatBoost 20 iterations, 5 binary targets, Intel i7-11700K:

| Metric | Rebuild Pool | `set_label` | Improvement |
|---|---|---|---|
| Pool construction (5 targets) | 5.2 s | 1.0 s | 5× |
| Full pipeline (5 targets) | 8.3 s | 3.2 s | 2.6× |
| Prediction match | — | Bit-exact on CPU, `rtol=1e-4` on GPU | preserved |

Script: `catboost/benchmarks/set_label_benchmark/benchmarks.py`. Pass `--scale-bench` for 100 k / 1 M / 10 M row throughput numbers.

## API contract

**Accepts**: `list`, `numpy.ndarray`, `pandas.Series`, 1-col `pandas.DataFrame`, `polars.Series`, 1-col `polars.DataFrame`. Numeric dtypes only. Existing Pool target types `None` / `Boolean` / `Integer` / `Float`.

**Rejects with clear errors**: non-numeric dtypes (`kind in {'O','U','S','V'}`), 2-D arrays with >1 column, length mismatch, Pools with existing `String` target.

**Dtype**: storage is `float32`; `get_label()` casts back to the input dtype. Integer values above `2**24` lose precision.

**Non-finite values** (`NaN`, `±Inf`) are accepted — mirrors the Pool constructor. Training losses reject them where relevant.

**Thread safety**: not safe to call concurrently with `fit`, `score`, or `eval_metrics` on the same Pool.

**Pitfalls**:
1. After `set_label` on a training Pool, update any `eval_set` Pool holding the prior labels. The dtype-mismatch warning does not catch same-dtype-but-stale-values.
2. `model.get_object_importance`, `get_feature_importance(type='LossFunctionChange')`, `eval_metrics(pool)`, and `calc_feature_statistics(pool)` re-read labels from the Pool at call time. If labels were mutated post-fit, those APIs use the new values with a model trained on the old ones. Pass a fresh Pool for introspection.

**Recommended patterns**:
- Multi-target: build the Pool once, then loop `pool.set_label(Y[:, i]); model.fit(pool)`. sklearn's `MultiOutputRegressor` does not preserve Pool metadata (its `check_X_y` coerces `X`) — use the plain loop.
- Hyperparameter search: construct the Pool outside the objective, call `pool.set_label(y_trial)` inside.

## Tests

**135 tests** in `catboost/python-package/ut/medium/test_set_label.py` — 132 pass, 3 GPU tests auto-skip without a CUDA device. Coverage:

- Prediction equivalence vs fresh-Pool construction: regression / binary / multiclass / ranking (YetiRank, QueryRMSE, PairLogit) / weighted / quantized / baseline / cat_features.
- Pool constructed with every accepted target type (`int64`/`int32`/`bool`/Python int list) + `set_label`; TSV-loaded string-labeled Pool rejected.
- Input types: numpy, Python list, pandas/polars Series & DataFrame.
- Error cases: non-numeric dtypes, 2-D multi-column, length mismatch (incl. ±1), `String` existing target.
- Edge cases: zero-row Pool, NaN/±Inf round-trip, negative integers, `int64` above `2**24` precision loss, refit-on-same-Pool bit-exact vs fresh baseline.
- Behavior-change coverage: `UserWarning` on overwrite in `fit` and all 3 `score` sites, `stacklevel` points at the user frame, `eval_set` aliasing raises across 5 shapes, strong exception guarantee on rejected input, dtype-mismatch warning.
- Training modes: `eval_fraction`, `cv()`, `eval_metrics`, `init_model` + inconsistent labels, Optuna-style loop, save/load round-trip.
- Backward-compat: classic `fit(pool)` path emits zero new warnings (promoted to errors in the test).

## Documentation

- `catboost/docs/en/concepts/python-reference_pool_set_label.md`
- `catboost/docs/en/_includes/work_src/reusage-python/set_label__desc.md`
- `catboost/docs/en/concepts/python-reference_pool.md`
- `catboost/docs/en/_includes/work_src/reusage/python__fit__y__note__main.md`
- `RELEASE.md`

## Checklist

- [x] 135 tests, bit-exact prediction equivalence verified across all task types.
- [x] Public API consistent with existing `Pool.set_*` methods.
- [x] Behavior changes documented in `RELEASE.md` and visible at runtime via `UserWarning`.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,3 +1,13 @@
+# Release (upcoming)
+
+## New features
+* \[Python-package\] Add `Pool.set_label(label)` to replace a Pool's labels in place without rebuilding the Pool. Enables efficient multi-target / weighting-schema sweeps on the same feature matrix at production scale.
+
+## Behavior changes
+* \[Python-package\] `model.fit(pool, y=labels)` previously raised `CatBoostError`; it now routes through `Pool.set_label(y)`. A `UserWarning` is emitted when this replaces pre-existing labels, so the mutation is not silent. The three `model.score(pool, y=...)` methods (classifier, regressor, ranker) received the same change. Passing the same `Pool` as both `X` and `eval_set` (directly, inside a list, inside a `(X, y)` tuple, or inside a list of such tuples) with a non-`None` `y` now raises a clear error -- previously it silently mutated the train pool's labels and invalidated the hold-out.
+* \[Python-package\] After `pool.set_label(y)`, `pool.get_label()` returns values cast back to the input dtype (e.g. `int64` in, `int64` out), matching the Pool constructor's behavior. Code that relied on `get_label()` always returning float after any mutation will see the original dtype instead.
+* \[Python-package\] `pool.set_label([])` on a zero-row Pool is now a legal no-op, mirroring `Pool(data=[], label=[])`.
+
 # Node package Release 1.27.0
 (uses `catboostmodel` native libraries from the main CatBoost release v1.2.10)
 * Fix prediction of type `Probability` on CPUs that do not have SSE4 instruction set (that includes all ARM CPUs).
--- a/catboost/benchmarks/set_label_benchmark/benchmarks.py
+++ b/catboost/benchmarks/set_label_benchmark/benchmarks.py
@@ -1,0 +1,390 @@
+"""
+Benchmark: Pool.set_label() vs Pool reconstruction for multi-target training.
+
+Demonstrates that reusing a Pool with set_label() produces identical predictions
+to rebuilding the Pool from scratch, while being significantly faster.
+
+Usage:
+    python benchmark_set_label.py [--rows 100000] [--cols 200] [--targets 5] [--iters 20]
+"""
+
+import argparse
+import time
+import sys
+
+import numpy as np
+from catboost import Pool, CatBoostRegressor, CatBoostClassifier
+
+
+def has_set_label():
+    """Check if Pool.set_label() is available (patched CatBoost)."""
+    return hasattr(Pool, 'set_label') and callable(getattr(Pool, 'set_label'))
+
+
+def generate_data(n_rows, n_cols, n_targets, seed=42):
+    """Generate feature matrix and multiple target vectors."""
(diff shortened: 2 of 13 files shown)


fingoldo:
Никаких новостей по рассмотрению PR?

PR #3088, opened 2026-05-05
Fix memory management Bugs for `PyList_SetItem`

wr-web:
https://github.com/catboost/catboost/issues/3087

### DIFF
--- a/contrib/tools/cython/Cython/Utility/CythonFunction.c
+++ b/contrib/tools/cython/Cython/Utility/CythonFunction.c
@@ -523,7 +523,6 @@ __Pyx_CyFunction_get_is_coroutine_value(__pyx_CyFunctionObject *op) {
         PyList_SET_ITEM(fromlist, 0, marker);
 #else
         if (unlikely(PyList_SetItem(fromlist, 0, marker) < 0)) {
-            Py_DECREF(marker);
             Py_DECREF(fromlist);
             return NULL;
         }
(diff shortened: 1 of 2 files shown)


andrey-khropov:
Cython in `contrib` is a vendored version of [the original Cython](https://github.com/cython/cython). I suggest to submit PR to the upstream Cython project. The change should be [around here](https://github.com/cython/cython/blame/master/Cython/Utility/CythonFunction.c#L590).
We will copy this change from the upstream code after that.

Also, CatBoost does not use the vendored Cython in `contrib` by default right now and uses system Cython instead. The vendored version is left for now just in case if we need some custom Cython logic in the future.

wr-web:
> Cython in `contrib` is a vendored version of [the original Cython](https://github.com/cython/cython). I suggest to submit PR to the upstream Cython project. The change should be [around here](https://github.com/cython/cython/blame/master/Cython/Utility/CythonFunction.c#L590). We will copy this change from the upstream code after that.
> 
> Also, CatBoost does not use the vendored Cython in `contrib` by default right now and uses system Cython instead. The vendored version is left for now just in case if we need some custom Cython logic in the future.

https://github.com/cython/cython/blob/cc0a466ff2f2c6bd5de439710750ac94ab831a0b/Cython/Utility/CythonFunction.c#L600-L601

Thank you very much for your reply. I noticed that this issue has already been fixed in Cython, but the corresponding update or fix hasn't been applied to the current project yet. I think fixing this issue is still valuable.

PR #3093, opened 2026-05-13
perf(c_api): eliminate per-call heap allocations in CalcModelPrediction

DenisRaskovalov:
## Problem

`CalcModelPrediction` with `catFeaturesSize=0` — the common pattern for float-only models — makes two unnecessary heap allocations on every call:

1. `TVector<TConstArrayRef<float>> floatFeaturesVec(docCount)`
2. `TVector<TVector<TStringBuf>> catFeaturesVec(docCount, TVector(0))`

It then dispatches to `model->Calc()`, the path designed for models with categorical features, paying for cat-feature bookkeeping that is never used when `catFeaturesSize==0`.

`CalcModelPredictionFlat` (multi-doc path) also allocates a fresh `TVector<TConstArrayRef<float>>(docCount)` on every call, even though the vector is trivially rebuilt from the caller's `float**` array.

### 1. Fast-path `CalcModelPrediction` when `catFeaturesSize == 0`

```cpp
if (catFeaturesSize == 0) {
    return CalcModelPredictionFlatStaged(
        modelHandle, docCount, treeStart, treeEnd,
        floatFeatures, floatFeaturesSize,
        result, resultSize);
}
```

When the caller passes `catFeaturesSize=0`, `CalcModelPredictionStaged` now delegates directly to `CalcModelPredictionFlatStaged`, routing through `model->CalcFlat()` instead of `model->Calc()`. This skips the `catFeaturesVec` allocation, the cat-feature loop, and the `Calc()` dispatch overhead.

### 2. Thread-local scratch buffer in `CalcModelPredictionFlatStaged`

```cpp
struct TFlatInferenceScratch {
    TVector<TConstArrayRef<float>> FloatFeatureRefs;
};
Y_STATIC_THREAD(TFlatInferenceScratch) FlatInferenceScratch;
```

The `TVector<TConstArrayRef<float>> featuresVec(docCount)` that was allocated fresh on every multi-doc call is replaced by a thread-local vector (using `Y_STATIC_THREAD`, the same mechanism as `ErrorMessageHolder`). The vector grows to the high-water mark of `docCount` and is reused across calls on the same thread — zero allocations after the first call per thread.

The `docCount==1` path already called `CalcFlatSingle` with no allocation; that fast path is preserved and documented.

## Safety

- **Thread safety**: `Y_STATIC_THREAD` gives each thread its own buffer. The CatBoost evaluator is stateless per call; concurrent threads sharing a `ModelCalcerHandle` are already supported and tested.
- **Correctness**: `CalcFlat` and `Calc` produce identical results when `catFeaturesSize==0`. Verified by the new test suite.
- **No API change**: all function signatures are unchanged.

## Tests added

`catboost/libs/model_interface/ut/c_api_flat_inference_ut.cpp` (5 test cases using `cloudness_small.cbm`):

| Test | What it checks |
|------|---------------|
| `NoCatFeaturesMatchesFlat` | `CalcModelPrediction(cat=0)` produces identical results to `CalcModelPredictionFlat` |
| `FlatMatchesStagedFullRange` | `CalcModelPredictionFlat` == `CalcModelPredictionFlatStaged(0, treeCount)` |
| `SingleDocMatchesBatch` | `docCount=1` fast path matches the batch path for the same input |
| `ConcurrentCallsAreConsistent` | 8 threads × 50 iterations — no data races on the thread-local buffer |
| `RepeatedCallsConsistent` | Alternating batch sizes (1, 10, 3, 50, 1, 7) verify buffer-resize correctness |

### DIFF
--- a/catboost/libs/model_interface/c_api.cpp
+++ b/catboost/libs/model_interface/c_api.cpp
@@ -36,6 +36,14 @@ struct TErrorMessageHolder {
 
 Y_STATIC_THREAD(TErrorMessageHolder) ErrorMessageHolder;
 
+// Thread-local scratch buffers for CalcModelPredictionFlat.
+// Allocated once per thread, grown on demand, never freed until thread exit.
+// Eliminates per-call heap allocations in the hot inference path.
+struct TFlatInferenceScratch {
+    TVector<TConstArrayRef<float>> FloatFeatureRefs; // [docCount]
+};
+Y_STATIC_THREAD(TFlatInferenceScratch) FlatInferenceScratch;
+
 
 class TFeaturesDataWrapper {
 public:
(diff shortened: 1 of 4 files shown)


DenisRaskovalov:
Friendly ping 🙂 — this PR is mergeable (no conflicts, CI clean). Could a maintainer please take a look and merge when convenient? Happy to address any feedback. Thanks!

PR #3097, opened 2026-05-19
Fix sklearn.base.clone for CatBoost estimators with mutable params

### LINKED ISSUE #3078: [Python] sklearn.base.clone() fails with RuntimeError for CatBoost >= 1.2.10 when mutable params are set
# [Python] `sklearn.base.clone()` fails with `RuntimeError` for CatBoost >= 1.2.10 when mutable params are set

## Environment

- catboost: 1.2.10
- scikit-learn: 1.5.0 (also reproduces on 1.8.0)
- Python: 3.11, 3.12
- OS: Linux (x86_64)

## Description

`sklearn.base.clone()` raises a `RuntimeError` for any CatBoost estimator (`CatBoostClassifier`, `CatBoostRegressor`, `CatBoostRanker`) when a mutable constructor parameter such as `cat_features`, `text_features`, or `embedding_features` is provided.

This started with CatBoost 1.2.10, which changed `get_params` to copy mutable parameters (related: #2991). Prior versions (e.g., 1.2.8 and earlier) stored the original reference, so `clone()` worked without issues. `sklearn.base.clone` checks **object identity** (`param1 is param2`) after round-tripping through `get_params() → constructor → get_params()`. Since CatBoost 1.2.10 now returns a copy, the identity check fails.

This breaks any sklearn utility that relies on `clone()`, including `cross_val_score`, `GridSearchCV`, `RandomizedSearchCV`, and third-party tools like Optuna.

## Reproduction

```python
from sklearn.base import clone
from catboost import CatBoostClassifier, CatBoostRegressor, CatBoostRanker

# Works — no mutable params
clone(CatBoostClassifier())

# Fails — cat_features is copied internally, breaking clone's identity check
clone(CatBoostClassifier(cat_features=[0, 1]))
```

### Output

```
OK: clone(CatBoostClassifier())
OK: clone(CatBoostRegressor())
OK: clone(CatBoostRanker())
FAIL: clone(CatBoostClassifier(cat_features=[0, 1]))
      Cannot clone object CatBoostClassifier(cat_features=[0, 1]), as the constructor either does not set or modifies parameter cat_features
FAIL: clone(CatBoostRegressor(cat_features=[0, 1]))
      Cannot clone object CatBoostRegressor(cat_features=[0, 1], loss_function='RMSE'), as the constructor either does not set or modifies parameter cat_features
FAIL: clone(CatBoostRanker(cat_features=[0, 1]))
      Cannot clone object CatBoostRanker(cat_features=[0, 1], loss_function='YetiRank'), as the constructor either does not set or modifies parameter cat_features
```

## Root cause

In CatBoost's constructor, mutable parameters like `cat_features` are stored as a **copy** (e.g., `list(cat_features)`). When `sklearn.base.clone` calls `get_params()`, constructs a new instance, and then calls `get_params()` again, it compares the returned values with `is`. The copy breaks this identity check:

```python
# Inside sklearn.base._clone_parametrized (simplified):
new_object_params = new_object.get_params(deep=False)
for name in new_object_params:
    if new_object_params[name] is not init_params[name]:
        raise RuntimeError(...)  # <-- fails here for cat_features
```

## Suggested fix

Add a `__sklearn_clone__` method to CatBoost's base estimator class. This is sklearn's official extension point (available since scikit-learn 1.3) — when defined, `clone()` calls it instead of the default identity-checking logic:

```python
import copy

# In CatBoost's base estimator class:
def __sklearn_clone__(self):
    klass = self.__class__
    params = self.get_params(deep=False)
    new_params = {k: copy.deepcopy(v) for k, v in params.items()}
    new_obj = klass(**new_params)

    # Preserve sklearn metadata routing requests (set_fit_request, set_score_request, etc.)
    if hasattr(self, "_metadata_request"):
        new_obj._metadata_request = copy.deepcopy(self._metadata_request)

    return new_obj
```

This is generic (handles all current and future mutable params), minimal, and does not require changes to `__init__`, `get_params`, or `set_params`.

Note: the `_metadata_request` preservation is important because `clone()` constructs a fresh instance that loses any metadata routing configuration set via `set_fit_request()`, `set_score_request()`, etc. For example, `CatBoostRanker` requires `set_fit_request(group_id=True)` for sklearn's metadata routing to pass `group_id` during cross-validation. Without preserving this state, cloned estimators fail with `UnsetMetadataPassedError` when used inside `cross_val_score`, pipelines, or hyperparameter search.

stohirov:
get_params() returns deepcopy(self._init_params), so sklearn.base.clone's default `is`-identity round-trip check fails for any estimator constructed with a mutable param (cat_features, text_features, embedding_features, ...). This broke cross_val_score, GridSearchCV, Optuna, and any other tool relying on clone().

Add __sklearn_clone__ to _CatBoostBase (sklearn >= 1.3 hook) so cloning bypasses the identity check while still producing an independent estimator. Also preserve _metadata_request so set_fit_request(group_id=True) on CatBoostRanker survives the clone (otherwise sklearn's metadata routing raises UnsetMetadataPassedError inside cross_val_score / pipelines).

Fixes #3078

### DIFF
--- a/catboost/python-package/catboost/core.py
+++ b/catboost/python-package/catboost/core.py
@@ -1814,6 +1814,12 @@ class _CatBoostBase(object):
         model.__setstate__(state)
         return model
 
+    def __sklearn_clone__(self):
+        new_obj = type(self)(**{k: deepcopy(v) for k, v in self.get_params(deep=False).items()})
+        if hasattr(self, "_metadata_request"):
+            new_obj._metadata_request = deepcopy(self._metadata_request)
+        return new_obj
+
     def __eq__(self, other):
         return self._is_comparable_to(other) and self._object == other._object
 
--- a/catboost/python-package/ut/medium/test_with_sklearn.py
+++ b/catboost/python-package/ut/medium/test_with_sklearn.py
@@ -338,3 +338,45 @@ def test_sklearn_tags_nan_mode_forbidden(task_type):
     else:
         tags = model._get_tags()
         assert tags['allow_nan'] is False
+
+
+@pytest.mark.parametrize(
+    'estimator_cls',
+    [CatBoostClassifier, CatBoostRegressor, CatBoostRanker],
+    ids=['Classifier', 'Regressor', 'Ranker'],
+)
+def test_clone_with_mutable_params(estimator_cls):
+    for kwargs in (
+        {'cat_features': [0, 1]},
+        {'text_features': [0]},
+        {'embedding_features': [0]},
+    ):
+        original = estimator_cls(iterations=2, verbose=False, **kwargs)
+        cloned = sklearn.base.clone(original)
+        assert type(cloned) is estimator_cls
+        for key, value in kwargs.items():
+            assert cloned.get_params(deep=False)[key] == value
+
+
+def test_clone_without_mutable_params():
+    for cls in (CatBoostClassifier, CatBoostRegressor, CatBoostRanker):
+        sklearn.base.clone(cls(iterations=2, verbose=False))
+
+
+@pytest.mark.skipif(
+    scikit_learn_version < (1, 3),
+    reason="metadata routing requires scikit-learn >= 1.3"
+)
+def test_clone_preserves_metadata_routing_request():
+    original = CatBoostRanker(iterations=2, verbose=False).set_fit_request(group_id=True)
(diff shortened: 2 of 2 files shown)


stohirov:
@andrey-khropov can you check the pipelines please?

andrey-khropov (approved):

PR #3100, opened 2026-05-28
Fix unchecked Skip() return values in zero-copy model loading paths

dephnor:
IInputStream::Skip() may return fewer bytes than requested. Several zero-copy LoadNonOwning paths ignored Skip() return values, allowing a malformed model file with a forged size field to advance the stream past its buffer. Subsequent reads from the exhausted stream access uninitialized memory.

**Files changed (4 files, +8/-4 lines):**

- `TFullModel::InitNonOwning` (model.cpp) — check `in.Skip(coreSize)` return
- `TCtrValueTable::LoadThin` (ctr_value_table.cpp) — check `in->Skip(len)` return
- `TTextProcessingCollection::LoadNonOwning` — defense-in-depth Skip check
- `TEmbeddingProcessingCollection::LoadNonOwning` — defense-in-depth Skip check

The default (owning) `TFullModel::Load` path is protected by `LoadOrFail`. The zero-copy `InitNonOwning` / `ReadZeroCopyModel` path (used by C API and JVM bindings) was vulnerable.

All fixes use the same pattern as existing error handling in the codebase (`CB_ENSURE` with a descriptive message).

### DIFF
--- a/catboost/libs/model/ctr_value_table.cpp
+++ b/catboost/libs/model/ctr_value_table.cpp
@@ -73,7 +73,8 @@ void TCtrValueTable::LoadSolid(void* buf, size_t length) {
 void TCtrValueTable::LoadThin(TMemoryInput* in) {
     auto len = LoadSize(in);
     auto ptr = in->Buf();
-    in->Skip(len);
+    const size_t skipped = in->Skip(len);
+    CB_ENSURE(skipped == len, "Failed to read CTR value table in zero-copy model loading");
 
     using namespace  flatbuffers;
     Impl = TThinTable();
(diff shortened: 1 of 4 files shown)


andrey-khropov (approved):

andrey-khropov:
Thank you for the fix.

A couple of notes:

1) Using `Skip` in this way for any `IInputStream` is not correct because it might skip less bytes than requested even if the end of stream is not reached. The right way would be to call this function repeatedly until either the requested amount of bytes is skipped or the end of the stream is reached similar to how `LoadOrFail` does it with reading. We will add a function similar to `LoadOrFail` but for skipping to `IInputStream` API in the future.

However for `TMemoryInput` in particular `Skip` always either skips the amount of data requested or reaches the end of stream so your fix is correct.

2) We will add `[[nodiscard]]` attribute to `Skip` and similar functions in `IInputStream` API to prevent such bugs in the future.

Issue #3103, opened 2026-05-30
`load_model` from `blob` or `stream` ignores the Poisson link function and returns `RawFormulaVal`


i-d-lytvynenko:
**Description**
When loading a saved model trained with a `Poisson` loss function using `load_model(blob=...)` or `load_model(stream=...)`, the model silently loses its default prediction link function (exponentiation).

As a result, calling `predict()` on the stream-loaded model returns the logarithmic `RawFormulaVal` instead of the expected target space value. Loading the exact same model via a file path correctly returns the exponentiated predictions.

**To Reproduce**

```python
import os
import tempfile

import numpy as np
import pandas as pd
from catboost import CatBoostRegressor

original_path = "data/model"

test_data = pd.DataFrame(
    {
        ...
    }
)

print("1. LOAD FROM FILE PATH:")
model_from_path = CatBoostRegressor()
model_from_path.load_model(original_path)
pred_from_path = model_from_path.predict(test_data)
print(f"   Prediction: {pred_from_path[0]:.6f}")

print("\n2. LOAD FROM FILE PATH (copied model):")
with tempfile.NamedTemporaryFile(suffix=".cbm", delete=False) as tmp:
    # Save and reload to ensure it's the same
    model_from_path.save_model(tmp.name)
    tmp_path = tmp.name

model_from_file = CatBoostRegressor()
model_from_file.load_model(tmp_path)
pred_from_file = model_from_file.predict(test_data)
print(f"   Prediction: {pred_from_file[0]:.6f}")

print("\n3. LOAD FROM BLOB:")
with open(tmp_path, "rb") as f:
    model_bytes = f.read()
model_from_blob = CatBoostRegressor()
model_from_blob.load_model(blob=model_bytes)
pred_from_blob = model_from_blob.predict(test_data)
print(f"   Prediction: {pred_from_blob[0]:.6f}")

print("\n4. LOAD FROM STREAM:")
model_from_stream = CatBoostRegressor()
with open(tmp_path, "rb") as f:
    f.seek(0)
    model_from_stream.load_model(stream=f, format="cbm")
pred_from_stream = model_from_stream.predict(test_data)
print(f"   Prediction: {pred_from_stream[0]:.6f}")


os.unlink(tmp_path)

print(f"\nlog(from_path): {np.log(pred_from_path[0]):.6f}")
```

**Expected behavior**
[...]

PR #3105, opened 2026-06-03
Fix default prediction link function and parameter sync for stream/blob loaded models

AceMeistr:
**Fixed default prediction link function and parameter sync for stream/blob loaded models**

from [ISSUE #3103](https://github.com/catboost/catboost/issues/3103#issue-4555673322)

Changes Made:

**Change 1:**
Implement BUG change: `# TODO(ilyzhin) change on get_all_params after MLTOOLS-4758`
with code: 
```cpp
 loss_function = self._object._get_loss_function_name() if self.is_fitted() else None
```
now allows the model to directly refer the C++ model engine avoiding the Temporary Python Dictionary preventing parameter override

**Change 2:**
Implemented Parameter Sync issue

By implementing immediate copying of tthe configuration parameters to python dictionary when loaded from memory.
we make sure that ``get_param('loss_function')`` always returns `Poisson` rather than `NONE` 

with code changes at: 
**L1846**:
```
      def _sync_params_from_model(self):
        self._init_params = {}
        for key, value in iteritems(self._get_params_from_model_updated_with_init_params()):
            self._init_params[key] = value
        self._canonized_params = None
```

L (**1976, 1988, 1993**):  `self._sync_params_from_model()`
this change push the model to sync all params from model at each stage to ensure accurate config params

**ALL TEST CASES and C.MAKE reviews PASSED**

### DIFF
--- a/catboost/python-package/catboost/core.py
+++ b/catboost/python-package/catboost/core.py
@@ -6345,9 +6345,7 @@ class CatBoostRegressor(CatBoost):
                                 "RMSE, MultiRMSE, SurvivalAft, MAE, Quantile, LogLinQuantile, Poisson, MAPE, Lq, RMSPE or a custom objective object".format(loss_function))
 
     def _get_default_prediction_type(self):
-        # TODO(ilyzhin) change on get_all_params after MLTOOLS-4758
-        params = self._get_canonized_params()
-        loss_function = params.get('loss_function')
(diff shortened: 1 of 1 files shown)


AceMeistr:
@andrey-khropov

 Please Review this PR and provide review 

Thank You.

andrey-khropov (approved):

Issue #1103, opened 2019-12-03
Implement ColumnDescriptionClass


Noxoomo:
This class could be used instead of cd file https://catboost.ai/docs/concepts/input-data_column-descfile.html when creating Pool from filez. The class should have init function, methods load and save, and Pool init method should be able to use object of this class instead of cd file during initialization.

michiboo:
Hi Can I have a go at this issue?

theNilanjan:
Hi, I would like to work on this issue. Is it still available for contribution?

andrey-khropov:
> Hi, I would like to work on this issue. Is it still available for contribution?

Yes. I've assigned this issue to you.

PR #3109, opened 2026-06-06
Add ColumnDescription support for Pool initialization

theNilanjan:
## Summary

This PR adds support for using a `ColumnDescription` object instead of a column description file path when creating a `Pool` from a file.

### Changes

* Added `ColumnDescription` class with:

  * `__init__`
  * `load`
  * `save`
* Extended column description type validation to accept `ColumnDescription` objects.
* Updated Pool file-reading logic to handle `ColumnDescription` instances by saving them to a temporary `.cd` file before passing them to the existing backend.
* Updated related type checks and documentation.

### Example

```python
cd = ColumnDescription()
cd.load("train.cd")

pool = Pool(
    "train.tsv",
    column_description=cd
)
```

This keeps backward compatibility with existing string and `os.PathLike` column description inputs.

Related to #1103


catboost/python-package/catboost/core.py
```diff
@@ -599,6 +600,17 @@ def plot_features_selection_loss_graphs(summary):
...
+    def __init__(self):
+        self.lines = []
+
+    def load(self, path):
+        with open(path, 'r') as f:
+            self.lines = f.readlines()
+
+    def save(self, path):
+        with open(path, 'w') as f:
+            f.writelines(self.lines)
```
andrey-khropov, line 613: Incorrect indentation

catboost/python-package/catboost/core.py
```diff
@@ -882,13 +894,15 @@ def _check_delimiter(self, delimiter):
...
-            raise CatBoostError("Invalid column_description type={}: must be str or os.PathLike.".format(type(column_description)))
+            
+def _check_column_description_type(self, column_description):
+    """
+    Check type of column_description parameter.
+    """
+    if not isinstance(column_description, PATH_TYPES + (ColumnDescription,)):
+        raise CatBoostError(
+            "Invalid column_description type={}: must be str, os.PathLike or ColumnDescription.".format(type(column_description))
+        )
```
andrey-khropov, line 905: Incorrect indentation

catboost/python-package/catboost/core.py
```diff
@@ -1355,10 +1369,21 @@ def _read(
...
+            mode='w',
+            suffix='.cd',
+            delete=False
+        )
+        tmp.close()
+
+        column_description.save(tmp.name)
+        column_description = tmp.name
+
+    self._check_column_description_type(column_description)
```
andrey-khropov, line 1386: Incorrect indentation

andrey-khropov (changes requested):
What is the use-case for this?

A dedicated ColumnsDescription class can be beneficial but it should contain a structured typed information about columns, not just lines from a file.

theNilanjan:
Thanks for the feedback.

My initial implementation focused on providing an in-memory replacement for
a column description file with minimal changes to the existing workflow. I
agree that a dedicated "ColumnDescription" class would be more useful if it
exposed structured column metadata rather than simply storing raw file
contents.

I will rework the design to represent column descriptions as typed data and
provide serialization/deserialization through "load()" and "save()". This
would make the class useful both for constructing column descriptions
programmatically and for reading existing ".cd" files.

PR #3125, opened 2026-07-07
Fix get_object_importance compatibility error for MultiClass models

Aadityasharma1-programmer:
Fixes a bug where `get_object_importance` fails on multidimensional models (like `MultiClass`) with an `RMSE` incompatibility error. 

The error occurred because `loss_function` strings (e.g. `loss_function="MultiClass"`) were silently failing to parse in `ReadTJsonValue`, causing the metric to mistakenly default to `RMSE`. 

Added a fallback in `data_providers.cpp`, `model.cpp`, and `loss_description.cpp` to correctly parse raw string loss functions via `ParseLossDescription`. Included `test_catboost.py` to verify the fix.

### DIFF
--- a/catboost/libs/model/model.cpp
+++ b/catboost/libs/model/model.cpp
@@ -1376,7 +1376,12 @@ static TMaybe<NCatboostOptions::TLossDescription> GetLossDescription(const TFull
     TMaybe<NCatboostOptions::TLossDescription> lossDescription;
     if (model.ModelInfo.contains("loss_function")) {
         lossDescription.ConstructInPlace();
-        lossDescription->Load(ReadTJsonValue(model.ModelInfo.at("loss_function")));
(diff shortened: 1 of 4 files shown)


andrey-khropov (changes requested):
- Loss description parsing is fine as is, it is not a problem. The issue is proper adaptation of the current object importance algorithm for the multidimensional case.

- Regarding tests:
  Current object importance tests should be extended for `MultiClass` case.
  [Tests for CLI](https://github.com/catboost/catboost/blob/master/catboost/pytest/test.py#L6914-L6925)
  [Tests for Python package](https://github.com/catboost/catboost/blob/cda35a76cd70193c648bd296bb19f70730d9580f/catboost/python-package/ut/medium/test.py#L5109)

PR #3134, opened 2026-07-15
Add feature_names_in_ attribute to estimators (#3129)

### LINKED ISSUE #3129: Add `feature_names_in_` estimated attribute to CatBoost estimators
https://scikit-learn-enhancement-proposals.readthedocs.io/en/latest/slep007/proposal.html

stohirov:
Closes #3129

### DIFF
--- a/catboost/python-package/catboost/core.py
+++ b/catboost/python-package/catboost/core.py
@@ -1772,6 +1772,11 @@ class _CatBoostBase(object):
         params_str = ", ".join(f"{key}={val!r}" for key, val in sorted(self._init_params.items()))
         return f"{self.__class__.__name__}({params_str})"
 
+    def __getattr__(self, name):
+        if (name == 'feature_names_in_') and self.is_fitted():
+            return np.array(self._object._get_feature_names(), dtype=object)
+        raise AttributeError("'{}' object has no attribute '{}'".format(type(self).__name__, name))
+
     def __getstate__(self):
         params = self._init_params.copy()
         test_evals = self._object._get_test_evals()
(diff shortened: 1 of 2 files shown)


andrey-khropov (changes requested):
1) The proper way to get feature names is to use `self._object._get_feature_names()` similarly how it is done [here](https://github.com/catboost/catboost/blob/f7f3cb02611b1932c1927bffbbeeabf9d0c930ea/catboost/python-package/catboost/core.py#L2111) but the difference with `feature_names_` should be that if the model is not fitted there should be an error for `feature_names_in_`, not `None` return value. So, I propose to implement `__getattr__` for `_CatBoostBase` class that will return value for this attribute if the model is fitted or raise `AttributeError` otherwise.

2) Please add checks for `feature_names_in_` to the already existing tests that check `feature_names_` in `catboost/python-package/ut/medium/test.py` instead of adding new tests in `test_with_sklearn.py`.

catboost/python-package/catboost/core.py
```diff
@@ -1772,6 +1772,11 @@ def __repr__(self) -> str:
         params_str = ", ".join(f"{key}={val!r}" for key, val in sorted(self._init_params.items()))
         return f"{self.__class__.__name__}({params_str})"
 
+    def __getattr__(self, name):
```
andrey-khropov: Please add a type annotation (`str`) for the `name` argument.

andrey-khropov (changes requested):

andrey-khropov (approved):

PR #3138, opened 2026-07-16
Validate leaf_values count when loading oblivious trees from JSON

### LINKED ISSUE #3137: JSON model loader silently accepts malformed leaf_values (count not a multiple of 2^depth), producing a wrong ApproxDimension
### Summary

When loading a model from JSON, `GetObliviousModelTrees()` computes the approx dimension with an unchecked integer division:

```cpp
// catboost/libs/model/model_export/json_model_helpers.cpp  (GetObliviousModelTrees)
modelTrees->SetApproxDimension(value["leaf_values"].GetArray().ysize() / (1uLL << treeSize));
```

For a valid oblivious tree of depth `d`, `leaf_values` must contain exactly `approxDimension * 2^d` values. There is no check that the array size is actually a positive multiple of `2^d`. If the JSON has been hand-edited, produced by a buggy generator, or truncated in transit so that `leaf_values.size()` is **not** a multiple of `2^d`, the integer division silently rounds down and yields an `ApproxDimension` that is too small (or zero). The malformed model is loaded without error and later produces wrong predictions / out-of-bounds leaf reads at apply time, instead of failing fast at load.

### Steps to reproduce

1. Train and export any model to JSON (`ExportModel(model, "model.json", EModelType::Json)`).
2. Remove a single entry from a tree's `leaf_values` array so the count is no longer a multiple of `2^depth`.
3. Load it back with `ReadModel("model.json", EModelType::Json)`.

**Expected:** load fails with a clear error, because the tree is structurally invalid.
**Actual:** load succeeds; the resulting model has a truncated `ApproxDimension` and yields incorrect results / OOB leaf reads on apply.

### Scope / honesty note

This is a **data-integrity / robustness issue in the JSON loader**, not a remotely exploitable memory-safety vulnerability. A model produced by CatBoost's own exporter always satisfies the invariant; triggering it requires an intentionally or accidentally corrupted JSON. The intent is fail-fast validation at load time rather than silently loading a corrupted model.

### Proposed fix

Add a bounds check in `GetObliviousModelTrees()` before computing the dimension:

```cpp
const ui64 leafValuesCount = value["leaf_values"].GetArray().ysize();
const ui64 leavesPerDimension = 1uLL << treeSize;
CB_ENSURE(
    leafValuesCount > 0 && leafValuesCount % leavesPerDimension == 0,
    "Invalid oblivious tree in JSON model: leaf_values count (" << leafValuesCount
        << ") must be a positive multiple of 2^depth (" << leavesPerDimension
        << ") for tree of depth " << treeSize);
modelTrees->SetApproxDimension(leafValuesCount / leavesPerDimension);
```

This raises `TCatBoostException` on malformed input.

### Verification (already done locally)

- Added unit test `TestMalformedLeafValuesCountRejected` to `catboost/libs/model/ut/json_model_export_ut.cpp` (export → drop one leaf → assert `ReadModel` throws `TCatBoostException`).
- Full `model_ut` suite: 36/36 pass with the fix (including the new test); zero regressions across the other test files in the suite.
- Verified the new test **fails** on the unpatched loader (`ReadModel ... did not throw any exception`), confirming the check is what catches the bug.

I have a branch ready with the fix + test and am happy to open a PR referencing this issue if the maintainers agree with the approach.

FuturMix:
### What

Fixes #3137.

`GetObliviousModelTrees()` in `catboost/libs/model/model_export/json_model_helpers.cpp` computed the approx dimension via an unchecked integer division:

```cpp
modelTrees->SetApproxDimension(value["leaf_values"].GetArray().ysize() / (1uLL << treeSize));
```

When a JSON model's `leaf_values` count is not a positive multiple of `2^depth`, the division silently truncates and produces a wrong `ApproxDimension`. The malformed model then loads without any error and can lead to out-of-bounds leaf reads at apply time, instead of failing fast at load time.

### Change

- Add a `CB_ENSURE` guard in `GetObliviousModelTrees()` so a `leaf_values` count that is not a positive multiple of `2^depth` is rejected at load with a clear message.
- Add a unit test `TestMalformedLeafValuesCountRejected` in `catboost/libs/model/ut/json_model_export_ut.cpp` that exports a valid model, drops one leaf value so the count is no longer a multiple of `2^depth`, and asserts the loader raises `TCatBoostException`.

### Scope

This is a data-integrity / robustness fix for local JSON model loading — it makes a malformed model fail fast with a clear error rather than loading silently. It is **not** a remotely exploitable memory-safety issue.

### Verification

- Full `model_ut` suite passes locally (36/36).
- The new test fails on the unpatched loader (silent acceptance) and passes with the guard in place (red → green).

### DIFF
--- a/catboost/libs/model/model_export/json_model_helpers.cpp
+++ b/catboost/libs/model/model_export/json_model_helpers.cpp
@@ -378,7 +378,14 @@ static void GetObliviousModelTrees(const TJsonValue& jsonValue, TModelTrees* mod
         }
         int treeSize = value["splits"].GetArray().ysize();
         modelTrees->AddTreeSize(treeSize);
-        modelTrees->SetApproxDimension(value["leaf_values"].GetArray().ysize() / (1uLL << treeSize));
+        const ui64 leafValuesCount = value["leaf_values"].GetArray().ysize();
+        const ui64 leavesPerDimension = 1uLL << treeSize;
+        CB_ENSURE(
+            leafValuesCount > 0 && leafValuesCount % leavesPerDimension == 0,
+            "Invalid oblivious tree in JSON model: leaf_values count (" << leafValuesCount
+                << ") must be a positive multiple of 2^depth (" << leavesPerDimension
+                << ") for tree of depth " << treeSize);
+        modelTrees->SetApproxDimension(leafValuesCount / leavesPerDimension);
         for (const auto& split: value["splits"].GetArray()) {
             modelTrees->AddTreeSplit(split["split_index"].GetInteger());
         }
--- a/catboost/libs/model/ut/json_model_export_ut.cpp
+++ b/catboost/libs/model/ut/json_model_export_ut.cpp
@@ -2,7 +2,11 @@
 
 #include <catboost/libs/model/model_export/model_exporter.h>
 
+#include <library/cpp/json/json_reader.h>
+#include <library/cpp/json/json_writer.h>
 #include <library/cpp/testing/unittest/registar.h>
+
+#include <util/stream/file.h>
 
 using namespace std;
 using namespace NCB;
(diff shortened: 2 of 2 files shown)


catboost/libs/model/model_export/json_model_helpers.cpp
```diff
@@ -378,7 +378,14 @@ static void GetObliviousModelTrees(const TJsonValue& jsonValue, TModelTrees* mod
         }
         int treeSize = value["splits"].GetArray().ysize();
         modelTrees->AddTreeSize(treeSize);
-        modelTrees->SetApproxDimension(value["leaf_values"].GetArray().ysize() / (1uLL << treeSize));
+        const ui64 leafValuesCount = value["leaf_values"].GetArray().ysize();
```
andrey-khropov, line 381: `ysize()` is for signed size, please just use `size()`.

andrey-khropov (changes requested):

andrey-khropov:
Tests in CI failed not because of the changes in this PR but because an incorrect unstable version of `plotly` has been installed due to the bug in the CI code. Cherry-picking https://github.com/catboost/catboost/commit/e10520bb8ce6b13b1264baa902fef2c021697232 to the branch will fix this issue.

Issue #1645, opened 2021-04-13
Add support for pandas nullable types


freddyaboulton:
Problem: Catboost does not support the [nullable pandas datatypes](https://pandas.pydata.org/docs/user_guide/integer_na.html). As as a user I would like to build a data processing using the latest pandas features and then pass my data to catboost.  Thank you!
catboost version: 0.24.4. Pandas version 1.2.3.
Operating System: OSX
CPU: N/A
# GPU:

Repro

```python
from catboost import CatBoostClassifier
import pandas as pd
import pytest

for dtype in ['Int64', 'Float64', 'boolean']:
    
    cb = CatBoostClassifier()

    # ValueError if only features use nullable types
    df = pd.DataFrame({"a": pd.Series([True, False, True, False], dtype=dtype),
                       "b": pd.Series([4, 5, 6])})
    y = pd.Series([1, 0, 1, 0])
    
    with pytest.raises(TypeError):
        cb.fit(df, y)
    
    # ValueError if only target uses nullable types
    df = pd.DataFrame({"a": pd.Series([True, False, True, False])})
    y = pd.Series([1, 0, 1, 0], dtype=dtype)
    
    with pytest.raises(TypeError):
        cb.fit(df, y)

# No ValueError if use "old" types
for dtype in ['int', 'float', 'bool']:
    
    df = pd.DataFrame({"a": pd.Series([True, False, True, False], dtype=dtype)})
    y = pd.Series([1, 0, 1, 0])

    cb = CatBoostClassifier()
    cb.fit(df, y)
```

tamargrey:
Seems like everything but works as of catboost 1.1.1 (with support coming over the course of a few prior releases):

```python
    import pandas as pd
    from catboost import CatBoostClassifier

    for dtype in ['Int64', 'Float64', 'boolean']:
        
        cb = CatBoostClassifier()

        # ValueError if only features use nullable types
        df = pd.DataFrame({"a": pd.Series([True, False, True, False], dtype=dtype),
                        "b": pd.Series([4, 5, 6])})
        y = pd.Series([1, 0, 1, 0])
        
        cb.fit(df, y)
        
        # ValueError if only target uses nullable types
        df = pd.DataFrame({"a": pd.Series([True, False, True, False])})
        y = pd.Series([1, 0, 1, 0], dtype=dtype)
        
        cb.fit(df, y)
        
```

andrey-khropov:
[...]

PR #3142, opened 2026-07-19
Support pandas nullable dtypes and pandas.NA/NaT as missing values in Python package

### LINKED ISSUE #2037: Add support for new pandas.NA as a missing value
Problem:
New Pandas versions use pandas.NA instead of np.nan to denote missing values.
Currently catboost fails on such features:
```
TypeError                                 Traceback (most recent call last)
_catboost.pyx in _catboost._FloatOrNan()

TypeError: float() argument must be a string or a number, not 'NAType'

During handling of the above exception, another exception occurred:

TypeError                                 Traceback (most recent call last)
_catboost.pyx in _catboost.get_float_feature()
```
catboost version:
1.0.4
Operating System:
all
CPU:
all
GPU:
all

SergeevVladislav:
Fixes #2037, related to #1645.

## Problem

pandas nullable extension dtypes (`Int8`..`Int64`, `UInt8`..`UInt64`, `Float32`, `Float64`, `boolean`, `string`) represent missing values with `pandas.NA` instead of `np.nan`. CatBoost failed on such data with cryptic errors:

```
TypeError: float() argument must be a string or a number, not 'NAType'
```

This happened for numeric features (e.g. `boolean` columns with `pd.NA`, object-dtype columns containing `pd.NA`, list input containing `pd.NA`) and for labels with nullable dtypes.

## Changes

**`catboost/python-package/catboost/_catboost.pyx`**
- `_FloatOrNan`: treat `pandas.NA` as NaN. This fixes all generic input paths: object-dtype DataFrame columns, list-of-lists input, `pandas.Series` as `data`, embedding features. `pandas.NaT` is intentionally **not** treated as a missing value: it is designed for missing datetime data only, so it remains an invalid input and raises an error, as before.
- `_set_features_order_data_pd_data_frame`: vectorized conversion of pandas nullable numeric/boolean extension dtypes to `float32` with NaN (`to_numpy(dtype=np.float32, na_value=np.nan)`) instead of slow per-element conversion. The `to_numpy()` conversion is now performed inside each feature-type branch with its own logic.

**`catboost/python-package/catboost/core.py`**
- `_label_if_pandas_to_numpy`: labels that are `pandas.Series` with nullable numeric/boolean dtypes are converted to `float64` with NaN. Nullable labels without missing values now work for both classification and regression; labels with missing values produce a clear `CatBoostError` from the C++ core (missing target values are not supported by CatBoost, same as `np.nan` before).

## Backward compatibility

- No behavior changes for previously working inputs. All numeric features are stored as `float32` internally, so the vectorized conversion is value-identical to the previous per-element path.
- `pandas.NA`/`pandas.NaT` in categorical features remain unsupported and fail with the same errors as before this change (missing values in categorical features are not supported — same semantics as `np.nan` before).
- Guards for old pandas versions: `pandas.NA` is looked up with `getattr` (it exists only since pandas 1.0.0); nullable floating dtypes exist only since pandas 1.2.0.

## Tests

Added 12 test functions (54 with parametrization) to `catboost/python-package/ut/medium/test.py`:
- all nullable numeric/boolean dtypes as features (Pool contents + fit + predict)
- equivalence of models trained on nullable data and on `float64` data with `np.nan`
- `pandas.NA` in object-dtype columns, list input and `pandas.Series` input
- `pandas.NaT` in a numerical column raises an error (behavior preserved)
- nullable labels with and without missing values
- NA values in categorical/text features raise errors

Tested with pandas 3.0.5 and pandas 2.3.3: all new tests pass, and the broader pool/pandas/label/dataframe-related test slice passes with no regressions.

### DIFF
--- a/catboost/python-package/catboost/_catboost.pyx
+++ b/catboost/python-package/catboost/_catboost.pyx
@@ -51,6 +51,18 @@ except ImportError:
 
 _polars_is_series_of_list_to_numpy_correct = _polars_version >= (1, 9)
 
+# pandas.NA singleton is available only since pandas 1.0.0
+_pandas_NA = getattr(pd, 'NA', None)
+_pandas_NaT = getattr(pd, 'NaT', None)
+
+# names of pandas nullable extension dtypes for numeric and boolean data
+_pandas_nullable_numeric_dtype_names = frozenset([
+    'Int8', 'Int16', 'Int32', 'Int64',
+    'UInt8', 'UInt16', 'UInt32', 'UInt64',
+    'Float32', 'Float64',
+    'boolean'
+])
+
 np.import_array()
 
 cimport cython
@@ -1284,6 +1296,8 @@ cdef inline float _FloatOrNan(object obj) except *:
     cdef type obj_type = type(obj)
     if obj is None:
         return _FLOAT_NAN
+    elif obj is _pandas_NA or obj is _pandas_NaT:
+        return _FLOAT_NAN
     elif obj_type is float:
         return <float>obj
     elif obj_type is str or obj_type is unicode or obj_type is bytes or obj_type is _npbytes_ or obj_type is _npunicode_ or isinstance(obj, string_types + (_npbytes_, _npunicode_)):
@@ -2184,6 +2198,8 @@ cdef inline get_id_object_bytes_string_representation(
             # for some reason Cython does not allow assignment to dereferenced pointer, so use this trick instead
             bytes_string_buf_representation[0] = to_arcadia_string(id_object)
         else:
+            if id_object is _pandas_NA or id_object is _pandas_NaT:
+                raise CatBoostError("bad object for id: {}".format(id_object))
             if isnan(id_object) or int(id_object) != id_object:
                 raise CatBoostError("bad object for id: {}".format(id_object))
             bytes_string_buf_representation[0] = ToString[i64](int(id_object))
@@ -3186,6 +3202,10 @@ cdef object _set_features_order_data_pd_data_frame(
                     new_data_holders
                 )
             else:
+                if column_data.dtype.name in _pandas_nullable_numeric_dtype_names:
+                    # pandas nullable numeric/boolean extension dtypes:
+                    # missing values (pandas.NA) are converted to NaN in a vectorized way
+                    column_values = column_data.to_numpy(dtype=np.float32, na_value=np.nan)
                 new_data_holders += create_num_factor_data(
                     flat_feature_idx,
                     column_values,
--- a/catboost/python-package/catboost/_catboost.pyx
+++ b/catboost/python-package/catboost/_catboost.pyx
@@ -51,6 +51,17 @@ except ImportError:
 
 _polars_is_series_of_list_to_numpy_correct = _polars_version >= (1, 9)
 
+# pandas.NA singleton is available only since pandas 1.0.0
+_pandas_NA = getattr(pd, 'NA', None)
+
+# names of pandas nullable extension dtypes for numeric and boolean data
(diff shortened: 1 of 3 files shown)


catboost/python-package/catboost/_catboost.pyx
```diff
@@ -2184,6 +2198,8 @@ cdef inline get_id_object_bytes_string_representation(
             # for some reason Cython does not allow assignment to dereferenced pointer, so use this trick instead
             bytes_string_buf_representation[0] = to_arcadia_string(id_object)
         else:
+            if id_object is _pandas_NA or id_object is _pandas_NaT:
+                raise CatBoostError("bad object for id: {}".format(id_object))
```
andrey-khropov: I don't think special handling is needed here. 
The check below is only to prevent silent rounding of floating-point like types.
All unspecified unsuitable types will just fail at the conversion to `int` (including `pandas.NA` and `pandas.NaT`).

catboost/python-package/catboost/_catboost.pyx
```diff
@@ -1284,6 +1296,8 @@ cdef inline float _FloatOrNan(object obj) except *:
     cdef type obj_type = type(obj)
     if obj is None:
         return _FLOAT_NAN
+    elif obj is _pandas_NA or obj is _pandas_NaT:
+        return _FLOAT_NAN
```
andrey-khropov: `pandas.NaT` is not a valid value to represent a missing value in a numerical column, it is designed for missing datetime data only and should be treated as an invalid input to this function. So, the current behaviour for it should be preserved.

catboost/python-package/catboost/_catboost.pyx
```diff
@@ -3186,6 +3202,10 @@ cdef object _set_features_order_data_pd_data_frame(
                     new_data_holders
                 )
             else:
+                if column_data.dtype.name in _pandas_nullable_numeric_dtype_names:
+                    # pandas nullable numeric/boolean extension dtypes:
+                    # missing values (pandas.NA) are converted to NaN in a vectorized way
+                    column_values = column_data.to_numpy(dtype=np.float32, na_value=np.nan)
```
andrey-khropov, line 3207: There's already a call `column_values = column_data.to_numpy()` above.
Double conversion will be very ineffective.

Let's change the logic that there be no single call `column_values = column_data.to_numpy(...)` for all types of features before 'if' / `elif` / `else`, move it to inside  `if` / `elif` / `else` for each type with its own logic.

andrey-khropov (changes requested):

catboost/python-package/catboost/_catboost.pyx
SergeevVladislav: Done — removed the special handling. `pandas.NA`/`pandas.NaT` in categorical features now fail at the conversion below like any other unsuitable type. Tests updated accordingly (they expect the same errors as before this PR).

catboost/python-package/catboost/_catboost.pyx
SergeevVladislav: Done — `pandas.NaT` is no longer treated as NaN, only `pandas.NA` is. `NaT` in a numerical column raises an error exactly as before this change. Added `test_pandas_nat_in_object_feature_fails` to lock this behavior.

catboost/python-package/catboost/_catboost.pyx
SergeevVladislav: Done — removed the common `column_values = column_data.to_numpy()` call. Each if/elif/else branch now performs its own conversion; nullable numeric/boolean dtypes use `to_numpy(dtype=np.float32, na_value=np.nan)` in the numeric branch only.

andrey-khropov (approved):

PR #3144, opened 2026-07-19
Do not define n_features_in_ until fit is called (#3004)

### LINKED ISSUE #3004: [scikit-learn compatibility] `n_features_in_` must not be defined until `fit` is called
https://scikit-learn-enhancement-proposals.readthedocs.io/en/latest/slep010/proposal.html#solution

stohirov:
Fixes #3004

### DIFF
--- a/catboost/python-package/catboost/core.py
+++ b/catboost/python-package/catboost/core.py
@@ -1772,6 +1772,11 @@ class _CatBoostBase(object):
         params_str = ", ".join(f"{key}={val!r}" for key, val in sorted(self._init_params.items()))
         return f"{self.__class__.__name__}({params_str})"
 
+    def __getattr__(self, name):
+        if (name == 'n_features_in_') and self.is_fitted():
+            return getattr(self, '_n_features_in')
+        raise AttributeError("'{}' object has no attribute '{}'".format(type(self).__name__, name))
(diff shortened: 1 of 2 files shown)


catboost/python-package/catboost/core.py
```diff
@@ -1773,6 +1773,8 @@ def __repr__(self) -> str:
         return f"{self.__class__.__name__}({params_str})"
 
     def __getattr__(self, name: str):
+        if (name == 'n_features_in_') and self.is_fitted():
+            return getattr(self, '_n_features_in')
         if (name == 'feature_names_in_') and self.is_fitted():
             return np.array(self._object._get_feature_names(), dtype=object)
```
andrey-khropov: Don't duplicate `is_fitted()` check. Check it in the outer if and then check `name` value if this check passes.

```py
if self.is_fitted():
    if name == 'n_features_in_':
      ...
    if name == 'feature_names_in_':
      ...
```

andrey-khropov (changes requested):

andrey-khropov (approved):
Ok, let's do it this way for now, but I will refactor the code later to avoid duplication of attribute names (likely through changing the implementation of `is_fitted`).

PR #3189, opened 2026-09-08
fix(python): include license file in built wheels (PEP 639)

### LINKED ISSUE #3118: Python wheels missing license file
Python wheels typically have a license file in their metadata section, but catboost wheels do not have that:
Image

Compare against lightgbm for example:
Image

webzuweb:
Fixes #3118 — CatBoost wheels published to PyPI do not carry the license file in their metadata section (unlike e.g. lightgbm), so tools like `pip` / `importlib.metadata` surface no license for the package.

## Change

- **`mk_wheel.py`**: extract `add_license_file()` and call it from `make_wheel()` to copy the repository `LICENSE` into the wheel's `.dist-info/licenses/LICENSE` directory (the PEP 639 location). `make_record()` already walks the whole wheel tree, so the license is automatically added to `RECORD` with its sha256.
- **`for_mk_wheel/catboost.dist-info/METADATA`**: add the `License-File: LICENSE` field alongside the existing `License: Apache License, Version 2.0`.
- **`ut/small/test_mk_wheel.py`**: unit tests for `add_license_file()` (copies the file; raises when `LICENSE` is missing) and `make_record()` (license appears in `RECORD` with a sha256).

## Verification

Local check of the new logic (copy + `make_record`):

```
license exists: True
RECORD has license: True
catboost-1.2.10.dist-info/licenses/LICENSE,sha256=-fjGX2iHnOt4E-yOAh9XplvtBwoLhrnpy42uDcRtTK8,28
```

Note: the `setup.py` sdist already bundles `LICENSE` (via `copy_catboost_sources`); this change addresses the wheel path built by `mk_wheel.py`.


andrey-khropov (changes requested):
CatBoost no longer uses `mk_wheel.py` to build wheels. 

Right now `pyproject.toml` and `setup.py` are used to build CatBoost python wheels.
See [the documentation](https://catboost.ai/docs/en/installation/python-installation-method-build-from-source)

webzuweb:
Thanks @andrey-khropov — reworked the fix to target the current build path (setup.py), dropping the mk_wheel.py changes entirely.

What changed:
- Reverted the `mk_wheel.py` / `for_mk_wheel/catboost.dist-info/METADATA` edits and removed `test_mk_wheel.py`.
- `setup.py` now passes `license_files=[os.path.relpath(os.path.join(get_topsrc_dir(), 'LICENSE'), SETUP_DIR)]` to `setup()`, so setuptools bundles the repo LICENSE into `.dist-info/licenses/`.
- The path is derived from the existing `get_topsrc_dir()` helper, which already resolves correctly both for source-tree builds and sdist builds (where LICENSE is copied into `catboost_all_src/`).
- Added `ut/small/test_setup_license.py` to cover the resolution logic.

Local verification (setuptools 84.0.0, the top of the pinned `>=70.1.0, <85.0` range): the built wheel contains `<pkg>.dist-info/licenses/LICENSE`.

PR #3198, opened 2026-09-18
Fix class weights in grid and randomized search

### LINKED ISSUE #2392: using `class_weights` with grid_search seems to provide incorrect results
catboost version: 1.2.0, Python package
Operating System: Windows 10

The **class_weights** parameter provides incorrect results with grid_search for CatBoostClassifier. In the case below, best estimator will always have the FIRST **class_weights** value that is specified in the *params* list. 
```py
params = {'class_weights': [[1, 1], [1, 2]]} 
cb = CatBoostClassifier(eval_metric='F1')
cb.grid_search(params, train)  # train is a pool object
```
Then `cb.get_all_params()` gives:
```py
...
'class_weights': [1, 1]
...
```
However, if we use `params = {'class_weights': [[1, 2], [1, 1]]}` to perform the grid search, then `cb.get_all_params()` will return:
```py
...
'class_weights': [1, 2]
...
```

The 'live' output of grid search also has only 3 folds for the first parameter, then it doesn't show any verbose stuff at all. Nor does the plot work.

I believe the root cause of all these issues lies within how the **class_weights** parameter is implemented in CatBoost. When we use it, the calculated metric is weighted (more on this [here](https://github.com/catboost/catboost/issues/1135) and [here](https://github.com/catboost/catboost/issues/2385)). That means, if we indicate `class_weights = [1, 1000]`, we'll get very good metrics. And grid search can't actually apply these metrics to compare models predictions, because it will always be biased towards the more imbalanced **class_weights** parameter. However, right now the grid search doesn't even compare the metrics for different values of **class_weights**, it just thinks that the first value in line is the best.

There are also a few open issues which seem to be connected with more or less the same problem: [1](https://github.com/catboost/catboost/issues/1888), [2](https://github.com/catboost/catboost/issues/2129), [3](https://github.com/catboost/catboost/issues/1275)

---
Some notes and questions and thoughts:

1) Is there any reason that CatBoost uses the weighted metric by default? From my perspective this doesn't seem to have any point and causes a lot of confusion (see all the mentioned issues). 
  
  
2) If I don't use grid_search, I can just specify: 
    ```py
    cb = CatBoostClassifier(eval_metric='F1:use_weights=false')
    ```
    This will eventually give me the unweighted metric that I want.  I thought I could fix the grid_search behaviour using the same 
    approach. However, if I do this:
    ```py
    params = {'class_weights': [[1, 1], [1, 2]]} 
    cb = CatBoostClassifier(eval_metric='F1:use_weights=false')
    cb.grid_search(params, train)
    ```
   I'll get the following error instead:
   ```py
   CatBoostError: C:/Go_Agent/pipelines/BuildMaster/catboost.git/catboost/libs/metrics/metric.cpp:6376: If non-default weights for 
   objects are not set, the 'use_weights' parameter must not be specified.
   ```
 
3) All of the above is also true for the **scale_pos_weight** parameter.
4) I have tried only the 'F1' metric. But it seems, according to the open issues, this is not the only metric affected.
5) My first and naive suggestion is to make the metric calculated by catboost *unweighted* by default. This may probably help to solve some of the issues and make CatBoost interface more user-friendly.

Thank you!

RAVasiliev:
Fixes #2392.

Grid and randomized search reused the first candidate's class weights, so changing their order could change the result. This updates the weights for each candidate while keeping the quantized features cached. It also fixes the `use_weights=false` error when class weights are set. Weighted metrics remain the default.

Tested on CPU/macOS ARM64: all 28 new regression cases fail before the fix and pass after it. All 12 existing search tests pass too.


RAVasiliev:
The macOS build failed during `conan install` because it couldn't resolve `center2.conan.io` ([log](https://github.com/catboost/catboost/actions/runs/35403819046/job/105960573256)). This PR doesn't change the build scripts or dependencies, and the Linux and Windows jobs passed. Could you please rerun the failed job?

catboost/python-package/ut/medium/test_search_class_weights.py
```diff
@@ -0,0 +1,82 @@
+import io
+
+import numpy as np
+import pytest
+from sklearn.datasets import make_classification
```
andrey-khropov: Tests that do not test compatibility with `sklearn` API should not depend on it.
Use [generate_random_labeled_dataset](https://github.com/catboost/catboost/blob/9a61977fbbc0138bb5ce1cc2876ef97dc4a27a8b/catboost/pytest/lib/common_helpers.py#L140) from catboost pytest lib instead.

catboost/python-package/ut/medium/test_search_class_weights.py
```diff
@@ -0,0 +1,82 @@
...
+
+from catboost import CatBoostClassifier, Pool
+
+
+def _search(pool, param_name, values, search_method, metric, train_test=True):
```
andrey-khropov: Please, add type annotations

catboost/python-package/ut/medium/test_search_class_weights.py
```diff
@@ -0,0 +1,82 @@
...
+@pytest.mark.parametrize('sample_weights', [False, True])
+@pytest.mark.parametrize('param_name,values,n_classes,data_kind', [
+    ('class_weights', [[1, 1], [1, 5]], 2, 'raw'),
+    ('auto_class_weights', ['None', 'Balanced'], 2, 'raw'),
+    ('class_weights', [[1, 1, 1], [1, 5, 2]], 3, 'raw'),
+    ('class_weights', [[1, 1], [1, 5]], 2, 'quantized'),
+    ('class_weights', [[1, 1], [1, 5]], 2, 'categorical'),
+])
+def test_search_class_weights_match_independent_candidates(
+    search_method, sample_weights, param_name, values, n_classes, data_kind
```
andrey-khropov: Please, add type annotations

catboost/python-package/ut/medium/test_search_class_weights.py
```diff
@@ -0,0 +1,82 @@
...
+        np.testing.assert_allclose(actual_score, expected_score, rtol=1e-10, atol=1e-12)
+
+
+@pytest.mark.parametrize('search_method', ['grid_search', 'randomized_search'])
+@pytest.mark.parametrize('train_test', [False, True])
+@pytest.mark.parametrize('param_name,values', [
+    ('class_weights', [[1, 1], [1, 5]]),
+    ('auto_class_weights', ['Balanced', 'SqrtBalanced']),
+])
+def test_search_class_weights_allow_unweighted_metric(search_method, train_test, param_name, values):
```
andrey-khropov: Please, add type annotations

andrey-khropov (changes requested):

